PR #3, opened 2017-04-02
Be able to remove some list parameters by emptying them

ldidry:
If you set an anonymous_sender, you can't remove it using the web
interface (not the only list parameter with this problem).
This allows to remove these parameters.

### DIFF
--- a/src/cgi/wwsympa.fcgi.in
+++ b/src/cgi/wwsympa.fcgi.in
@@ -11797,8 +11797,11 @@ sub do_edit_list {
                     }
                     ## Else if the parameter has only a scalar value
                 } else {
-                    ## Remove if empty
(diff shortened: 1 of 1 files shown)


ikedas:
Luc,
(Don't worry about failed travis check.)

I wonder whether your patch will fix the bug: In the line 11775 parameter is checked if its occurrence is "0-n" or "1-n", but delivery_time etc. seems to have occurrence "0-1".

ldidry:
You're right. The fact is that I needed to apply this patch (without the `[$i]`) on my production server (6.2.16): I didn't really checked what was different in the upstream code.

We can close this PR. Sorry for the noise.

ikedas:
I'll investigate this bug in my side.  Thank you for reminding!

Issue #10, opened 2017-04-21
(6.2.16) issues with include_sympa_list via a template?


agouaux:
I have the following definitions.

/deploy/sympa/etc/data_sources/owner_or_moderator_sympa_list.incl:

`include_sympa_list`
`name owner_or_moderator_sympa_list`
`listname [% param.0 %]`
 
/deploy/sympa/list_data/ecs.bmen.staff/config:

`owner_include`
`profile privileged`
`reception mail`
`source owner_or_moderator_sympa_list`
`visibility conceal`
`source_parameters etech`

However, even though the logic seems to work, get the following in the logs:

Apr 21 10:15:49 aispsympa01 task_manager[35865]: info Sympa::List::_load_include_admin_user_file() 
Bad entry "listname [% param.0 %]" for key "listname", paragraph "include_sympa_list" in 
/deploy/sympa/etc/data_sources/owner_or_moderator_sympa_list.incl

Apr 21 10:15:49 aispsympa01 task_manager[35865]: info Sympa::List::_load_include_admin_user_file() 
Missing key "listname" in param "include_sympa_list" in 
/deploy/sympa/etc/data_sources/owner_or_moderator_sympa_list.incl

agouaux:
I just discovered that some of these list config files had *_include paragraphs that were missing the `source_parameters X` line.  In some cases the list had two paragraphs, one with the `source_parameters` line, and another without it.

It turns out, that once my data source for including a list has been selected, if the list owner then wishes to unset that value, there's no way to de-select it.  So what happens is that value is selected, but they deleted the source parameter (the list name), hence the incomplete paragraph and the resulting error messages.

<img width="614" alt="screen shot 2017-04-21 at 12 47 17 pm" src="https://cloud.githubusercontent.com/assets/27364661/25289827/d920369e-2690-11e7-9461-94a1deed175b.png">

ikedas:
agouaux,

Latter of your reports already has been filed as a bug.
https://sourcesup.renater.fr/tracker/index.php?func=detail&aid=9972&group_id=23&atid=167

ikedas:
So it is a duplicate of the bug fixed after 6.2.16.  This issue is closed by now.

PR #13, opened 2017-05-03
[6.2] Rewriting code on list configuration form

ikedas:
Entirely refactored edit_list_request.tt2. Simpler and consistent logic to compile configuration

Fixed bugs:
* [SourceSup#9741] "percentage of list members in VERP mode" is order by left justified
* Access privilege ("read" etc.) of paragraph did not override that of sub-parameter:
  * e.g. If "archive" gives "read" privilege and "archive.web_access" gives "write", latter is editable on edit_list form but change is ignored.
* [SourceSup#7914] Web interface let define the same email twice as list owner.
* [SourceSup#9972] Existing item in member_include list configuration paragraph cannot be removed.
* Text parameter values including "%xx" are broken: For example "%a5" is altered to "/", "%3a" to ":".
* and several minor flaws.

Changes:
* List parameter "cookie" will be masked, even if it is accessible (normally by listmaster).
* Obsoleted "date" parameter of following paragraphs: last_instantiation, creation and update.
* Internal parameters, e.g. "creation", may not be editable.

Known bug:
* If owner_include, editors or editor_include is empty, then [Update] on edit_list form is clicked without any changes, configuration file will be updated.
This seems to be caused by inconsistency of current code to save/load config file, and may be fixed by another pr.

(6 May update)
(8 May update)

### DIFF
--- a/default/web_tt2/copy_template.tt2
+++ b/default/web_tt2/copy_template.tt2
@@ -24,8 +24,9 @@
 [%|loc%](This template is the default for all languages unless it is redefined for a specific language)[%END%]
 [%- ELSE -%]
 <strong class="neutral"
-lang="[%tpl_lang_lang%]" xml:lang="[%tpl_lang_lang%]"
-> [%tpl_lang_title%] </strong>
+ lang="[%tpl_lang_lang%]" xml:lang="[%tpl_lang_lang%]">
+[%~ tpl_lang | optdesc('lang',1) ~%]
+</strong>
 [%- END %] <br />
 
 </fieldset>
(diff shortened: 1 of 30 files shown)


ikedas:
Another bug fix:
* [SourceSup#7914] Web interface let define the same email twice as list owner.

ikedas:
One more bug fix:
* [SourceSup#9972] Existing item in member_include list configuration paragraph cannot be removed

This bug was also mentioned in Issue #10.

ikedas:
Known bug:
* If owner_include, editors or editor_include is empty, then [Update] on edit_list form is clicked without no changes, configuration file will be updated.

This seems to be caused by inconsistency of current code to save/load config file, and may be fixed by another pr.

Issue #22, opened 2017-06-09
include_ldap_query parameter attrs not valid (6.2.17b1)


MorenoUniTN:
Hi to all,
my problem is the **-** caracter in the filed attrs on the include_ldap_query.
In the last version of sympa (6.1.x) my workaround is changed the value of the **List.pm** at line:

'include_ldap_query' => {'format' => {'host' => {'format...
...
**'attrs' => {'format' => '\w+',**

to:

'include_ldap_query' => {'format' => {'host' => {'format...
...
**'attrs' => {'format' => '.+',**

but now???? This problem blocking my organization....

racke:
Sorry, I can't find this piece of code. Please either state the file and line number or include it in the issue as

```
code block
```

ikedas:
@MorenoUniTN, what lind of attribute name you want to enter? Could you show example?

ikedas:
@MorenoUniTN, does the change above fix the problem?

MorenoUniTN:
@racke This part of code is refered to fle List.pm of Sympa version 6.1 the new List.pm of Sympa 6.2 is complety differents and not contain this part of code

@ikedas Yes, in the old version of Sympa 6.1 this workaround fix the problem

there is a place in the new sympa 6.2 to set enable the caracter - in attrs parameter?
Thanks to all

ikedas:
I'll merge PR #23.   I feel appreciate if you check the next beta.

MorenoUniTN:
@ikedas Thanks for all, but my respons above is refered to old code of old version sympa 6.1.x
I try to modify file indicated at #23 but i have the same error.
The caracter is - dash in the attrs value of ldap query
![screenshot 2017-06-14 at 08 53 38](https://user-images.githubusercontent.com/29304925/27119188-29a311cc-50df-11e7-9beb-c1158624e70d.png)

ikedas:
@MorenoUniTN, can you apply this patch to new version?
https://github.com/sympa-community/sympa/pull/23.patch

MorenoUniTN:
@ikedas Great! now it's ok.
thanks a lot for all!!!

ikedas:
Good. So this issue will be closed.  Thank you for reporting bug!

PR #23, opened 2017-06-11
Issue #22: Parameters for LDAP attribute description are too restrictive

ikedas:
This PR is intended to be fix to Issue #22.

Parameters for LDAP attribute description such as "include_ldap_query" did not allow those including option(s) like "cn;lang-ja".  Fixed regular expression to check their format according to [RFC2251, 4.1.5](https://tools.ietf.org/html/rfc2251#section-4.1.5).


racke:
Good idea to use the regular expression from the RFC :+1:

PR #66, opened 2017-09-29
"%" sign in SQL data source would be escaped

ikedas:
[Submitted by B. Marchal, univ. Lorraine] When using '%' symbol in SQL data sources, said % was ignored because it was interpolated in subsequent sprintf. Fixed by escaping % sign.

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -5827,9 +5827,11 @@ sub _include_users_sql {
         = @_;
 
     my $sth;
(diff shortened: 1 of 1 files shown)


racke:
👍

ikedas:
Using do_prepared_query() seems preferred (See a diff above).

Reminder: Change is incompatible. Previously some people got over by duplicating "`%`" in sql_query parameter.  We should note this change in release note.

PR #67, opened 2017-09-29
Allow multiple attrs in include_ldap_2level_query

### LINKED ISSUE #63: include_ldap_query and include_ldap_2level_query behaving differently
While playing with include_ldap_query, I found what I think is an undocumented behaviour.
If filling attrs with, say,  "mail,cn", both the email and name appear when reviewing the list members.
The fact that you can use 2 attributes in the attrs field, first for email and second for name, is not documented.

However, the same behaviour cannot be reproduced with include_ldap_2level_query.
Using attrs2="mail" works, the dynamic list is properly populated with the emails.
Using attrs2="mail,cn" just gives an empty subscribers list.

This is with sympa 6.2.19 beta2.

xavierba:
fixes #63

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -5568,7 +5568,7 @@ sub _include_users_ldap_2level {
             'search',
             base   => "$suffix2",
             filter => "$filter2",
-            attrs  => ["$ldap_attrs2"],    # FIXME: multiple attrs?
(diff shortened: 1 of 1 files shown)


racke:
👍

dverdin:
I don't see any trouble here... Looks good to me. :+1:

PR #77, opened 2017-10-04
Deprecate sympa.spec and META.json.

ikedas:
These files bundled in source distribution would be deprecated:
- sympa.spec - Poorly maintained. Spec for Fedora/EPEL is separately developed by volunteers: \[[repo](https://github.com/ikedas/sympa-rpm)\].
- META.json - Never used and unusable. It would be better to be incorpolated into [Task::Sympa](https://metacpan.org/release/Task-Sympa).

My commit removes these files from distribution.


racke:
Both can go, I agree. Thanks!

PR #81, opened 2017-10-06
Refactoring requests more

ikedas:
This is a continuation of refactorings done by PR #26, #30, #33, #65.  In this PR, following handlers are introduced:
- `decl` to decline held requests.  On WWSympa, decl_add and decl_del (formerly ignoresub and ignoresig).
- `move_list` to move or copy list, formerly rename_list and copy_list on WWSympa.

And, "`sympa.pl --rename_list`" and "`sympa.pl --import`" also uses request handlers.

Fixed bugs:

- [bug] Content of several tables and spools were not updated when list was renamed: exclusion_table, notification_table, digest spool, topic spool, outgoing (bulk) spool.
- [bug] Couldn't rename/copy to list with the same name and other domaain.

- [bug] If `list_check_smtp` is disabled, a list can be renamed or copied to the list with system address, <sympa@_domain_> or <listmaster@_domain_>.

- [bug] If digest spool contains garbage, Sympa::Spool::Digest::Collection->next may crash.

Known bug:

- move_list handler is still unsolid: If it is executed during message delivery, some messages may be lost or renaming may abort.  Further improvements are desirable.


racke:
With a better database schema, list renames would be automatically reflected in the db by foreign keys/trigger.

Issue #80, opened 2017-10-05
Fix ridiculous English spelling for spam protection button


racke:
`This button trys to protect the mailing lists archives against address harvesting by a Spammer.`

3rd person of try is "tries" - and spammer should be lowercase. In addition this message is quite pathetic in general. I will *try* to come up with a better one.

ikedas:
Errors in translation (including English) would be corrected on translation site: https://translate.sympa.org/
Would you please check it?

racke:
OK, done that.

ikedas:
Closd at present.

Issue #54, opened 2017-09-08
Button label confusing on form to send HTML page 


racke:
The button on the view `compose_mail/eshop-news/html_news_letter` says _Send to selected receipient_, but it simply sends it to the whole list. So it is confusing (aside from the typo ...).

racke:
Fix looks good, thanks!

Issue #52, opened 2017-09-07
Text change / typo fix in change_email_request template


racke:
There is at typo in `default/web_tt2/change_email_request.tt2` (shoukd instead of should). However, I propose to change this sentence into "Please check your mailbox now". This version is more polite and user friendly in my opinion.

```
<!-- change_email_request.tt2 -->

[%|loc%]Changing your email address is a sensitive operation so we need to verify your email.[%END%]<br />
[%|loc(new_email)%]To this end we have sent you an email to this address: %1 with a validation link.[%END%]<br />
[%|loc%]You shoukd check your mailbox now.[%END%]

<!-- end change_email_request.tt2 -->
```

racke:
Thanks for the fix, looks good!

PR #84, opened 2017-10-07
When fallback language "en" is used, use "en_US" translation catalog

ikedas:
This change is related to issue #80 (and probably #52, #54).

When language context is not known, fallback setting `en` is used. Currently, it does not use translation catalog and there seems no easy way to correct artless messages.  This PR may solve the problem by trying to use `en_US` translation catalog for `en-US` also in `en` context: Corrections (at least minor ones) can be done modifying the translation.


racke:
Looks reasonable 👍

PR #120, opened 2017-11-11
Insignificant bugs related to cache

ikedas:
- When list is closed, list could be marked error_config. Fixed by checking owner only when list is open or peinding.
- When list is restored, long time could be taken before owners/editors are restored.


racke:
👍

PR #121, opened 2017-11-11
Fix listname typo

jean1:
list name was not rendered as a dot was missing
(listname instead of list.name)


ikedas:
In the case `type == 'hundred_percent_error'`, `list.name` is defined, but in the case `type == 'automatic_list_creation_failed'`, it is *not* defined. In latter case `listname` looks correct.

ikedas:
I merged your PR partly (a part of your changes are reverted).
This PR will be closed.  Thank you for reporting & fixing bug!

PR #123, opened 2017-11-17
Changes on POD format for transition to Markdown

ikedas:
At least with kramdown, a variant of Markdown, two hyphens (`--`) in the text is altered to an en dash: See <https://github.com/gettalong/kramdown/issues/472>.  To avoid it, use `C<...>` instead of `B<...>` etc. in POD.


ikedas:
Uncritical changes and will merged immediately.

PR #131, opened 2017-11-28
Feature: optionally restrict list ownership to specific domains (owner_domain)

mpkut:
Motivation
==========

Our campus mailing list service allows a wide range of uses by our institutional community.  Our default is to be open wherever possible, including the ability to share moderation and ownership tasks with people at other institutions. At the same time we are obliged to ensure that mailing list service resources are reserved for our community and not effectively given away to external parties. And we need to do that at a scale of thousands of lists and list owners.

We originally approached this need by modifying wwsympa.fcgi to implement a policy that can be stated as follows:

> **At least one list owner address must be in a known institutional domain.**

It was not a perfect approach but has worked well enough in practice.

With our recent work to migrate to Sympa 6.2, we have reimplemented this idea as a general purpose Sympa feature.

Description
===========

The list definition panel features two new per-list options:

* owner_domain: space separated list of required domains for owner addresses.

The default is an empty list, meaning no restrictions are applied.

* owner_domain_min: minimum number of owners required to match the owner_domain list.

If unset or 0, then all list owner addresses must match the owner_domain list.

The default editing permissions in edit_list.conf allow privileged owners to modify the owner_domain list, while restricting access to read-only for unprivileged owners.

The default values for owner_domain and owner_domain_min can be set site-wide in sympa.conf.

Implementation Notes
==================

The existing validation mechanism in Sympa::List::Config only provides a way to review individual configuration file items. Also, no validation function is called at all when a configuration item is removed. This is insufficient when implementing a policy that needs to take the whole updated configuration into account.

To address this, _validate_changes() now calls a set of "global" validation functions after all per-item validation functions are complete. The owner_domain/owner_domain_min check is implemented as a
global validation function.

Files
=============

src/lib/Sympa/List/Config.pm:

* set up per-list option configuration
* add ability to run global validation functions on entire updated config
* provide global validation function for owner_domain/owner_domain_min options

src/lib/Sympa/ListDef.pm:

* set up per-list config parameters

src/lib/Sympa/ConfDef.pm:

* set up sympa.conf config parameters

mail_tt2/report.tt2:

* present error messages when owner_domain or owner_domain_min requirements are not met while editing list owners

default/edit_list.conf:

* specify default editing privileges

Example 1
==========

An interest group with a unique mail domain wants to make sure that all list owners of a specific important list are associated with the group's domain.

1) Use the list definition panel to make sure that all list owners match the domain in question
2) Update the owner_domain setting for the list to read

```
groupdomain.org
```

Example 2
==========

A large institution has a policy that all lists across the entire mailing list service must maintain at least one owner address matching a list of known institutional domains.

1) Update sympa.conf as follows

```
owner_domain domain1.edu domain2.org domain3.net ...
owner_domain_min 1
```

2) Update edit_list.conf to make the settings read-only for all list owners, reserving write access to the listmaster:

```
owner_domain                    owner,privileged_owner          read
owner_domain_min                owner,privileged_owner          read
owner_domain                    listmaster                      write
owner_domain_min                listmaster                      write
```

Limitations
===========

* The current implementation only reviews the main list configuration file. It does not consult any other data sources used for owners.


mpkut:
A side comment. This feature feels a little specialized, I wlll admit. Hopefully we done our best to make the defaults and behavior as broadly suitable as possible, and perhaps attractive enough to consider for inclusion in the main line. Thank you for your review!

--mic--

ikedas:
Merged. Thanks!

mpkut:
Much appreciated!

PR #134, opened 2017-12-04
Characters in pages are garbled (get "mojibake") with Perl 5.22.0 or later

ikedas:
This bug was reported by a japanese listmaster.  Users in non-Latin-1 languages (language other than "en", "de", "fr", ...) will notice this bug. Latin-1 users rarely realize it but warnings are spit out to error log:
```
wwsympa.fcgi: Use of wide characters in FCGI::Stream::PRINT is deprecated and will stop wprking in a future version of FCGI at /usr/local/lib/perl5/site_perl/mach/5.24/Template.pm line 162.
```

Background: Starting with Perl 5.22.0, POSIX::strftime() returns Unicode (utf8 flag set) string under UTF-8 locale.
- https://metacpan.org/pod/release/SHAY/perl-5.26.1/pod/perl5220delta.pod#Better-heuristics-on-older-platforms-for-determining-locale-UTF-8ness
- https://perl5.git.perl.org/perl.git/commit/9717af6d049902fc887c412facb2d15e785ef1a4

Sympa::Language::gettext_strftime() uses it to format localized date/time, and utf8-flagged texts were mixed in the output.

Fixed by dropping utf8 flag from the result of POSIX::strftime().

----
4 Dec Update: Added warning message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
   - autoreconf -i
   - ./configure
   - cd src; make; cd ..
-  - make check-local TEST_FILES='t/compile_executables.t t/compile_modules.t t/parse_templates.t t/pod-syntax.t'
+  - make check-local TEST_FILES='t/compile_executables.t t/compile_modules.t t/parse_templates.t t/pod-syntax.t t/Language.t'
 
 after_success:
   - coverage-report
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
   - autoreconf -i
   - ./configure
   - cd src; make; cd ..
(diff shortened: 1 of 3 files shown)


racke:
In most cases it should not be necessary to mess with the UTF-8 flags, so maybe the real problem is elsewhere?

ikedas:
Well, this type of problem messes non-Latin-1 people.

Sympa uses binary UTF-8 (utf8 flag off) as internal text encoding, not Unicode (utf8 flag on). Today several modules return Unicode strings as results. Mixture of binary UTF-8 and Unicode strings will be "upgraded" to Latin-1 (with "wide characters" warnings).

So with Latin-1 people, problem will often not be actualized. However with non-Latin-1 people, things are serious: Most or all of non-ASCII letters will be broken.

PR #188, opened 2018-02-05
Pictures max size fix

### LINKED ISSUE #180: Hardcoded max picture size in picture_upload.tt2
Subscriber's max picture size defaults to 100 Kb, but can be changed with the pictures_max_size option. However, picture_upload.tt2 hardcodes the 100 Kb value.
https://github.com/sympa-community/sympa/blob/091ebf5b3949e834d30644f815932f76eb23c103/default/web_tt2/picture_upload.tt2#L9

xavierba:
This fixes #180, but I'm not completely satisfied with the fix, as the pictures_max_size will be properly reported but the size will be expressed in bytes, while it would probably be better to report it in kilobytes.

### DIFF
--- a/default/web_tt2/picture_upload.tt2
+++ b/default/web_tt2/picture_upload.tt2
@@ -6,7 +6,7 @@
 	 
  	 <form method="post" action="[% path_cgi %]" enctype="multipart/form-data" >
 
-<p>[%|loc(conf.pictures_max_size / 1024)%]You can upload your picture below. It will be available in the list review page. The picture should use a standard format (GIF, JPP, JPEG or PNG) and the file size should not exceed %1 Kb.[%END%]</p><br />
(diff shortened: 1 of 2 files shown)


xavierba:
I've corrected the 2nd commit to use kilobytes again. Thanks for the hint @ikedas :-)

racke:
Change looks good to me 👍

ikedas:
@xavierba, good fixes!

btw, though this question is not related to this fix, what is "JPP" format?

xavierba:
This is most probably a typo for JPG. See https://github.com/sympa-community/sympa/blob/407a17e4a8cea342c3657fe8243a6855a9621685/src/cgi/wwsympa.fcgi.in#L13919
I've added one more commit to fix it.

PR #196, opened 2018-02-16
Add symlink pointing from lib to src/lib so prove -vl works out of the box

racke:


dverdin:
I agree with this. It is a short term solution to enable prove commands without having to run make and configure.
Even though configure and make will be necessary as long as the Sympa modules remain intertwined.

ikedas:
I suppose we may use `PERL5LIB` environment variable.

In addition, I feel `FindBin` used in some test files is not required by same reason (to begin with, I feel it is no use to *.t files be location-independent).

What is the reason to avoid use of `PERL5LIB`?

racke:
@ikedas The upside of using and supporting prove -l is that you really use the modules in the current directory instead of whatever is in PERL5LIB. That prevents wondering why your changes doesn't reflect in the tests.

ikedas:
I thought of current build system.

I guess there may be a difficulty to use prove: Since perl specified by `--with-perl` configure option is used, another version of `prove` can be used according to `PATH` variable.
Contrary, if we stick to Makefile generated by autotools, `PERL5LIB` may always have sticked value --- as current Makefile does: https://git.io/vAclv

ikedas:
@racke,

How about moving `src/lib` to `lib`? It may be done easily and looks better than symlink for me.

racke:
> @racke,
> 
> How about moving `src/lib` to `lib`? It may be done easily and looks better than symlink for me.

Yes, that would be the most straightforward solution.

ikedas:
> > @racke,
> > How about moving `src/lib` to `lib`? It may be done easily and looks better than symlink for me.
> 
> Yes, that would be the most straightforward solution.

@racke, if you still think this idea is promising, please submit the new issue (and close this pr).

ikedas:
Stale discussion. I'll close it for now.  Re-opening is welcome.

Issue #195, opened 2018-02-16
Language test fails


racke:
```
not ok 91 - set_lang(cs-CZ-lasstina)
#   Failed test 'set_lang(cs-CZ-lasstina)'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'cs'
not ok 92 - set_lang(cs-lasstina): locale-independent case
#   Failed test 'set_lang(cs-lasstina): locale-independent case'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'cs'
not ok 93 - set_lang(cs-CZ)
#   Failed test 'set_lang(cs-CZ)'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'cs'
not ok 94 - set_lang(cs)
#   Failed test 'set_lang(cs)'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'cs'
ok 95 - set_lang(en-CA): no catalog (en_CA) - fallback to en
ok 96 - set_lang(en-Shaw): no catalog (en@shaw) - fallback to en
ok 97 - set_lang(en): no catalog (en) - fallback to en
not ok 98 - set_lang(zh): macrolanguage zh
#   Failed test 'set_lang(zh): macrolanguage zh'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'zh-TW'
not ok 99 - set_lang(zh-guoyu): macrolanguage zh
#   Failed test 'set_lang(zh-guoyu): macrolanguage zh'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'zh-TW'
not ok 100 - set_lang(zh-TW)
#   Failed test 'set_lang(zh-TW)'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'zh-TW'
not ok 101 - set_lang(zh-Hant): semi-macrolanguage zh-Hant
#   Failed test 'set_lang(zh-Hant): semi-macrolanguage zh-Hant'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'zh-TW'
not ok 102 - set_lang(zh-Hant-TW)
#   Failed test 'set_lang(zh-Hant-TW)'
#   at t/Language.t line 254.
#          got: undef
#     expected: 'zh-TW'
not ok 103 - set_lang(zh-Hant-HK): semi-macrolanguage zh-Hant
#   Failed test 'set_lang(zh-Hant-HK): semi-macrolanguage zh-Hant'
#   at t/Language.t line 254.
[...]

PR #199, opened 2018-02-17
#195: Some tests fail with checked out source

ikedas:
This may fix issue #195.

Sources checked out from repository lacks *.mo test data files so that some tests in t/Language.t fail. Only "make dist" generate these files.

Fixed by letting check-local rule depend on test files.


racke:
Thanks, this patch works and is pretty much straightforward. I suggest to merge it right away.

PR #203, opened 2018-02-20
Use standard 644 perms for $sysconfdir/README

xavierba:
As in title, not much else to say :-)

Regards,
Xavier


ikedas:
+1

444 looks not so harmful but I have no reason to force it.

racke:
Makes sense :+1:

PR #208, opened 2018-02-23
Sympa::Aliases re-writing with Sympatic.

dverdin:
OK, Here is where I am about Sympa::liases modules family.
I did the following:
1- Adding test files for all modules,
2- Adding them make check-local,
3- Replacing imports with Sympatic when aplicable,
4- Merging modules: got rid of Sympa::Aliases ::CheckMofule. Now an option sets whether aliases need to be used.
5- Removed module guessing from the main Sympa::Aliases module. Modules are explicitely named from calling programs. I know this creates duplicat code in callers and this is bad. But I want modules to be testable without loading Sympa config. that means trying to eradicate contextual behaviour. It is not ficnished and certainly not perfect but I think it is a step. Please let me know if you agree.

Id' like now to remove dependencies to config in order to not call sympa config inside these modules. Only passing them parameters.

I also want to find a good way to choose between modules depending on Sympa config. I still don't know where it should be.

Anyway, comments are welcome!


racke:
The subject of this PR is too generic, please adjust it so it is clear that we are talking about alias modules/scripts.

racke:
Make fails:
```

make[1]: Entering directory '/home/racke/sympa-community/sympa/src/lib'
make[1]: *** No rule to make target 'Sympa/Aliases/CheckSMTP.pm', needed by 'Sympa/Internals.pod'.  Stop.
make[1]: Leaving directory '/home/racke/sympa-community/sympa/src/lib'
Makefile:370: recipe for target 'all-recursive' failed
```

racke:
- Sympatic dependencies are missing.
- Implementation of "add" and "del" in subclasses should be enforced.

racke:
Please use isa_ok for testing where an object is created properly.

ikedas:
>> Following code block tends to deal with another alias plugin, e.g. if $type is "Foo", "Sympa::Aliases::Foo" subclass is loaded. It means that user can add "alias_manager Foo" line to sympa.conf so that they can use Sympa::Aliases::Foo plugin written by their own or prepared by default.
> 
> Actually, I kept that code but moved it in calling modules.
I'll do a PR today with my full work and you'll be able to comment on the full picture.
Thanks for all these usefull comments!

@dverdin, I expected that you will show how we can do what current code does using Moo, but you included non-trivial refactoring.  If you wish to do it, could you submit PR for sympa-6.2? I worry dispersed refactoring might invite mess.

I also suggest to use the new branch on your fork for PR: Working on sympa-7.0 on your fork may make retry difficult.

ikedas:
> Id' like now to remove dependencies to config in order to not call sympa config inside these modules. Only passing them parameters.

I think it is impossible.  anywhere we tried to put aside config processing, in the end some module must take it over (for example, you hate Sympa::Aliases::CheckSMTP depending on config, and moves it to Sympa::Aliases, then had to split it).

Instead, with ongoing refactoring on 6.2, I plan to implement the site config as a singleton instance.  To test any modules, we would load this singleton or alternative stub (see also my post on dev list at 1 Dec).

On the other hand, on current code, stub of Conf.pm is easily defined: We simply may define %Conf::Conf variable with appropriate content.

dverdin:
Dear all,
Thanks for your comments!
A few answers before I leave for two weeks:
- @Ikedas: I'm afraid to do refactoring in Sympa 6.2. It is the stable branch, I fear to break things...
- @Ikedas again: About config: I wouldn't say removing it from modules is impossible but it is certainly hard and not always relevant. I think we must do it for modules we want to distribute in CPAN. For the rest, using a Singleton is indeed the best solution.
- @Racke: I'll fix the bugs you noticed when I'm back. and I'll use "isa_ok". For now, I just wanted to check whether the modules coud be used and the new sub worked. But as you can see, test files are far from being complete. I'll work on them too with actual alias data files to be created by tests.

Anyway, I think it is a promising work and has the virtue to tackel the questions of how to transform Sympa using actual code, thus asking actual questions.

See you soon!

ikedas:
> Dear all,
> Thanks for your comments!
> A few answers before I leave for two weeks:
> 
> @ikedas: I'm afraid to do refactoring in Sympa 6.2. It is the stable branch, I fear to break things...
But Sympa::Aliases and children are born by refactoring during very recent 6.2 (6.2.23b). Stability and refactoring may coexist, I think.

If you fear, you may submit PR instead of changing on stable branch by yourself.

Honestly, I’m reluctant to begin development of 7 by now.  Sympa needs major surgery including refactoring — as you see, what is going on in 6.2 is merely rewriting. Our fault is  that we have put off such work and have told on the brilliant future.

Anyway, if you feel refactoring is required, I feel it would be tried on 6.2 at first.  How do you think?

racke:
On 02/26/2018 02:21 AM, IKEDA Soji wrote:
> *@ikedas* commented on this pull request.
> 
> ------------------------------------------------------------------------------------------------------------------------
> 
> In src/lib/Sympa/Aliases.pm <https://github.com/sympa-community/sympa/pull/208#discussion_r170477935>:
> 
>>      }
>  
> -    # Returns appropriate subclasses.
> 
> It seems you removed following code block and did not wirte alternative thing.
> 
>   * This loads plugin (a subclass of Sympa::Aliases) specified by type. It allows users to write plugin easily.
>   * Loading on demand descreases memory usage, and avoids possible dependency loop.
> 
> Similar patterns appear in several cases (Sympa::Spindle etc.). Could you please show us how you can implement such
> pattern using Moo?

I both cases I would recommend Module::Runtime :-)

Regards
          Racke


-- 
Ecommerce and Linux consulting + Perl and web application programming.
Debian and Sympa administration. Provisioning with Ansible.

ikedas:
As discussed and agreed on [IRC meeting at 6 April](http://sympa-ja.org/irc/freenode/sympa/2018/sympa.2018-04-06-07.38.html), we'll close this PR, and separation of 6.2 and 7.0 (either fork/branch/repo level) may be discussed further.

PR #222, opened 2018-03-06
Create all needed spool directories at install time

xavierba:
Hi,

Some spool directories are not created at install time.
The following directories are created:
- automatic
- bounce
- msg
- task
- tmp
The following directories are missing:
- bulk
- digest
- moderation
- outgoing
- subscribe
- viewmail

This PR fixes the creation of the missing directories.

Regards,
Xavier

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -115,7 +115,8 @@ installdir:
 		chown $(USER) $(DESTDIR)$$dir; \
 		chgrp $(GROUP) $(DESTDIR)$$dir; \
 	done
-	-@for subdir in automatic bounce msg task tmp; do \
+	-@for subdir in automatic bounce bulk digest moderation msg outgoing \
+		subscribe task tmp viewmail; do\
 		if [ ! -d $(DESTDIR)$(spooldir)/$$subdir ] ; then \
 			echo "Creating $(DESTDIR)$(spooldir)/$$subdir"; \
 			$(INSTALL) -d -m 750 $(DESTDIR)$(spooldir)/$$subdir; \


ikedas:
The first five spools are mandatory because they possiblly cannot be created when they are required: Three for incoming messages, task spool<sup>[1]</sup> and temporary workspace.  The others are used only by Sympa's component, and it is guaranteed that they will be created when they are required.

I suppose we don't worry just creating the five, even if the new spool will be added in the future.

<sup>[1]</sup> The task spool have to be created because Sympa won't create it at this time.  It may be fixed in near future.

xavierba:
Also missing:
bounce/bad/
msg/bad/

ikedas:
I investigated the source code of 6.2.30.

 - Following directories located under SPOOLDIR by default are created automatically by any of Sympa's processes (deamons, sympa.pl, wwsympa etc.) when the processes have to access.
 However, some of directories may be accessed by auxiliary programs which assume existence of corresponding directory but don't create it.

 | Description | Derectory (default location) | Auxiliary program(s) |
 |-------------|------------------------------|---|
 | held message spool | `$SPOOLDIR/auth` | - |
 | automatic list creation spool | `$SPOOLDIR/automatic` | [familyqueue(8), sympa-milter](https://sympa-community.github.io/manual/customize/automatic-lists.html) |
 | ,, (quarantined) | `$SPOOLDIR/automatic/bad` | - |
 | bounce incoming spool | `$SPOOLDIR/bounce` | bouncequeue(8) |
 | ,, (quarantined) | `$SPOOLDIR/bounce/bad` | - |
 | message outgoing spool | `$SPOOLDIR/bulk` | - |
 | ,, (quarantined) | `$SPOOLDIR/bulk/bad` | - |
 | ,, | `$SPOOLDIR/bulk/bad/msg` | - |
 | ,, | `$SPOOLDIR/bulk/bad/pct` | - |
 | ,, (messages) | `$SPOOLDIR/bulk/msg` | - |
 | ,, (recipient packets) | `$SPOOLDIR/bulk/pct` | - |
 | digest spool | `$SPOOLDIR/digest/` | - |
 | ,, | `$SPOOLDIR/digest/`_list id_ | - |
 | ,, (quarantined) | `$SPOOLDIR/digest/`_list id_`/bad` | - |
 | moderation spool | `$SPOOLDIR/moderation` | - |
 | message incoming spool | `$SPOOLDIR/msg` | queue(8) |
 | ,, (quarantined) | `$SPOOLDIR/msg/bad` | - |
 | archive spool | `$SPOOLDIR/outgoing` | - |
 | ,, (quarantined) | `$SPOOLDIR/outgoing/bad` | - |
 | ,, (migrated) | `$SPOOLDIR/outgoing/migrated` | - |
 | held request spool | `$SPOOLDIR/subscribe` | - |
 | topic cache | `$SPOOLDIR/topic` | - |
 | cache of formatted bounce messages | `$SPOOLDIR/viewmail/bounce` | - |
 | cache of formatted messages held | `$SPOOLDIR/viewmail/mod` | - |

 - Following directories seems not created by Sympa's processes nor auxiliary programs, but obviously necessary.

 | Description | Derectory (default location) | Auxiliary program(s) |
 |---|------------------------------|---|
 | task spool | `$SPOOLDIR/task` | - |
 | temporary directory | `$SPOOLDIR/tmp` | virus scanner etc. |

ikedas:
@xavierba,
After all, directories which must be created explicitly (otherwise Sympa won't work correctly) are these:

| Description | Derectory (default location) | Auxiliary program(s) |
| --- | --- | --- |
| automatic list creation spool | `$SPOOLDIR/automatic` | [familyqueue(8), sympa-milter](https://sympa-community.github.io/manual/customize/automatic-lists.html) |
| bounce incoming spool | `$SPOOLDIR/bounce` | bouncequeue(8) |
| message incoming spool | `$SPOOLDIR/msg` | queue(8) |
| temporary directory | `$SPOOLDIR/tmp` | virus scanner etc. |

Of course there may be some reason to create the other directories. What do you think?

dverdin:
I agree with @ikedas : we don't need to create spools until we need to use them.
I don't see why they should be created at install time.

xavierba:
The main reason is consistency. It seems more consistent to me that either none or all of the directories at the same depth are created at install time.
In this case, either sympa only creates $SPOOLDIR at install time and all $SPOOLDIR subdirectories are created at runtime, either sympa creates $SPOOLDIR and all $SPOOLDIR subdirectories at install time.
It would probably help avoid issues like this one: https://github.com/ikedas/sympa-rpm/pull/51

racke:
It looks like `X509-user-certs` wasn't considered in our directory layout.

ikedas:
I thought this PR focused on directories under $SPOOLDIR.  OTOH [`ssl_cert_dir`](https://sympa-community.github.io/manual/man/sympa.conf.5.html#ssl_cert_dir) is `$EXPLDIR/X509-user-certs` by default (some distribution assign other location as default).  Anyway I'll comment on it shortly.

(Despite its name) `ssl_cert_dir` is referred to by [S/MIME](https://sympa-community.github.io/manual/customize/smime.html) feature.  Since S/MIME feature is optional, existence of this directory could be optional (On ikedas/sympa-rpm#51, creation of directory for RPM package was proposed, not for upstream).  That's why I think creation of this directory at install time is not necessary.

ikedas:
> The main reason is consistency. It seems more consistent to me that either none or all of the directories at the same depth are created at install time.
> In this case, either sympa only creates $SPOOLDIR at install time and all $SPOOLDIR subdirectories are created at runtime, either sympa creates $SPOOLDIR and all $SPOOLDIR subdirectories at install time.
> This PR was a try at implementing the latter, but another way would be to search and fix all occurrences of spool dirs that are not automatically created when needed.

Why I support the former as much as possible is that default location of spools will not always be used.
  * Note that locations of all directories discussed here are configurable: See [sympa.conf(5)](https://sympa-community.github.io/manual/man/sympa.conf.5.html#directories).
    In fact, moving location of some directories are useful.  For example: Putting `msg` and `bulk` on SSD; Sharing `auth`, `moderation`, `subscribe` and so on with network drive by multiple hosts (under a load balancer) or DRBD volume (on HA cluster).
    In these cases, since system admins may create directories manually under other volumes, the default location may not be used.
  * Not all directories are required in all use cases.
    `auth`, `automatic`, `bounce`, `moderation`, `outgoing` (\*), `digest` or `topic` (and maybe more) will not be necessary, if corresponding feature was explicitly disabled or simply not used.

So I think $SPOOLDIR and minimum necessary subdirectories would be created at install time.

----
(\*) `outgoing` is used for archiving, not for handling outgoing messages.

dverdin:
I concur with @ikedas : For example, if you don't use the automatic lists, you don't need the automatic spool. So why create it?

ikedas:
> I concur with @ikedas : For example, if you don't use the automatic lists, you don't need the automatic spool. So why create it?

But I wrote `automatic` and `bounce` should be created: See above. So my proposal is certainly inconsistent.

xavierba:
I certainly agree that _all_ of the spool directories should be created only when needed. Eventually, sympa should only create the spool topdir on install, and all subdirectories should be created only when needed. Same for data dir.
I just don't like the inconsistency of some subdirectories are created at install time and some are not, but this is admittedly a bias of mine. Also, adding the capability to dynamically create directories would ease the life of admins changing the conf file to relocate some of them elsewhere.
Automatically creating directories is a much better solution, but as this is currently not possible, I tried to address it the other way round in this PR, that is create all dirs at install time.
Imho, this PR is certainly sub-optimal and shall be dropped, but the discussion that it sparkled is valuable.

ikedas:
This PR is closed by now.  Thanks for suggestive discussion!

Issue #11, opened 2017-04-21
Spurious error on duplicate keys with admin sync


agouaux:
I didn't know if this was something separate from #7 or something different.  #7 deals with editing the admins (owners and/or editors).  However, this problem has to do with syncing these admins from a datasource.  I'm going to attached an excerpt of the log regarding such a sync.  The lines with "AAG" are additional log statements I inserted in an attempt to get more information about this.  One thing that's frustrating is that I have not been able to replicate this behavior on the test system, despite them having the same sympa version, same perl version, same module versions (at least mostly.)

In the attached log excerpt, the list "ecs.bmen.staff" includes the list "etech" as owners.

[etech-sympa-log.txt](https://github.com/sympa-community/sympa/files/948112/etech-sympa-log.txt)

ikedas:
agouaux.

Could you please explain what you expected and how the result was unexpected?

As far as I understood: owners has already been registered in admin_table, and thanks to primary-key constraint of the table, duplicate registrations were avoided.

racke:
I think that this is caused by the stupid caching of admins and friends in 6.2.

agouaux:
I upgraded to 6.2.18, but the problem persists.

Here's an example.  List is called ecs.ee.allfac, and it includes the list etech for both the admins and editors.  That has been working fine for a long time.  However, after going to 6.2, now whenever I edit the List Definition, I get the following error message:

> User amos@x.org has encountered an internal server error
> (Web interface - ACTION: Command process - LIST:
> ecs.ee.allfac@lists.x.org):
> 
> Failed to include list admins for list 'ecs.ee.allfac'
> 
> See the logs for more details.

along with this nasty error message on the screen:

<img width="814" alt="screen shot 2017-07-10 at 10 11 51 am" src="https://user-images.githubusercontent.com/27364661/28025248-803a7434-6558-11e7-8e3e-be068e7e562d.png">

So I agree that on one hand it is a non-serious error since it is just trying to add a duplicate entry, but: 

* Either it needs to treat it as a non-serious error, or
* Add logic to determine if in fact those entries are already there and do not attempt to add duplicates.

Here's the log for when this occurs.  

[sympa.log.txt](https://github.com/sympa-community/sympa/files/1136021/sympa.log.txt)

agouaux:
Here's the list config, if that matters:

```
owner_include
profile privileged
reception mail
source owner_or_moderator_sympa_list
visibility conceal
source_parameters etech

editor_include
[...]

PR #223, opened 2018-03-06
Deprecate datetime field type in database

ikedas:
This PR deprecates datetime field type in database.  Because:
  - There are no portable way to compare, increase or decrease such field values.
  - It's slow: Values should be converted to Unix timestamp; Index cannot be applied.

Database will be upgraded automatically during upgrading process.

### DIFF
--- a/src/lib/Sympa/Database.pm
+++ b/src/lib/Sympa/Database.pm
@@ -473,22 +473,16 @@ sub quote {
 ## This sub takes a single argument: the name of the field to be used in
 ## the query.
 ##
-sub get_canonical_write_date {
-    my $self  = shift;
-    my $field = shift;
-    return $self->get_formatted_date({'mode' => 'write', 'target' => $field});
-}
+# Moved to Sympa::Upgrade::_get_canonical_write_date().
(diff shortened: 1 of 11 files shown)


racke:
I have some doubts about this idea ... we would have definitely less issues with it when we are using DBIx::Class. The other problem with using Unix timestamps is they are not legible on the SQL level compared to datetime entries.

ikedas:
> I have some doubts about this idea ... we would have definitely less issues with it when we are using DBIx::Class.

I agree to that we would be better to use smarter ORM.

> The other problem with using Unix timestamps is they are not legible on the SQL level compared to datetime entries.

Honestly, I wanted to use update_date_* fields to determine either each record have to be removed or not (this is a part of changes related to issue #11), and I found there is no portable way to compare between datetime field and system clock time (we have to support four or more dialects of SQL).

I guess these fields have been used to store just a time stamp and not considered arithmetic usage I wanted.

ikedas:
For example I want to do such: https://git.io/vAFnj

Anyway, if there are not strong objection, I'd like to merge my PR at this time.

Issue #233, opened 2018-03-14
Anyone can unsubscribe a member of a list with open scenario


qosobrin:
Debian Stretch 9.4 and Sympa Package 6.2.16 environment.

I have one list (list@lists.domain.org) configured with unsubscribe.open scenario:

```
title.gettext open

!is_subscriber ([listname],[email]) smtp,dkim,smime,md5 -> do_it
!equal ([sender],[email])           smtp,dkim           -> request_auth([email])
true()			            smtp,md5,smime      -> do_it
```

and three users: user_a@domain.org, user_b@domain.org and user_c@domain.org Both users A and B are members of the list, while user C is not.

When either user A (member) or user C (not member) sends a message to sympa@lists.domain.org with the following subject "SIGNOFF list@domain.org user_b@domain.org" then user B is automatically unsubscribed from the list. The expected behaviour should be user B to receive an authentication request, but it doesn't happen. 

Thank you very much. Regards.

ikedas:
Hi @qosobrin,

Can you also check recent version of Sympa?

qosobrin:
I have just checked with version 6.2.30 and the problem continues. When user A sends the SIGNOFF message to the list "SIGNOFF list@domain.org user_b@domain.org" User B is unsubscribed from the list. The funny thing is that User B receives the unsubscription mail notification.

qosobrin:
After editing the unsubscribe.open scenario and removing the ! sign in the second line:
```
title.gettext open

!is_subscriber ([listname],[email]) smtp,dkim,smime,md5 -> do_it
equal ([sender],[email])           smtp,dkim           -> request_auth([email])
true()			            smtp,md5,smime      -> do_it
```
I tried again and this time user B received the unsubscription confirmation request. It seems like either the equal scenario condition is returning the opposite value or the ! sign is not being  evaluated correctly.

ikedas:
Hi @qosobrin,

Can you show sympa log during original scenario is processed? (Log lines not including "`Sympa::Scenario::" are not needed.)

qosobrin:
Find attached the full log (debug 3) during user signoff ([issue_233.log](https://github.com/sympa-community/sympa/files/1861886/issue_233.log)). In this case I include both sympa and postfix logs. The particular lines referring to the scenario processing are these ones:
```
Mar 29 22:31:11 parasito sympa_msg[815]: debug2 Sympa::Scenario::_parse_scenario(tracking, default, listas.usc.es)
[...]

PR #246, opened 2018-04-02
[bug] unsubscribe.open scenario omits [email]

ikedas:
Fixed by assigning $request->{email} to scenario context.

This may fix issue #233.


racke:
That makes sense ... but your comment appears odd to me as we *add* the email address instead of suppressing it.

ikedas:
Thanks for correction.  I feel odd too. :)

racke:
:+1:

Issue #243, opened 2018-03-27
cookie parameter protection


xavierba:
Hi,
sympa.conf is set to mode 640 in order to protect the cookie from being world readable.
However, the sympa.conf.bin file is mode 644, whereas it should be 640 too.

Also, cookie should be marked as obfuscated in ConfDef.pm, in order to prevent it from being displayed in the web interface.

Regards,
Xavier

ikedas:
We worked on the same issue. :-)  Choose better one.

xavierba:
@racke didn't like my patch very much, and as I wrote I didn't even test it, I just wanted to update the issue. I'm not fond of the chown/chmod, it seems to me there is a race condition between file creation, chown and chmod steps, using umask seems slightly better, although not perfect. It would be even better to have the .bin files created with good owner/group and mode from the very start. Also, your PR is better, it handles the db_password parameter too.

ikedas:
ok @racke, could you please review these two when you come back (or submit the third by your own :))?

ikedas:
PR has been merged.

PR #248, opened 2018-04-04
Obfuscate cookie

xavierba:
Tentative fix for #243, currently untested.

### DIFF
--- a/src/lib/Conf.pm
+++ b/src/lib/Conf.pm
@@ -2368,6 +2368,18 @@ sub _save_binary_cache {
         }
         return undef;
     }
+    eval { chmod(0640, $param->{'target_file'}); };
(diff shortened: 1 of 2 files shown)


racke:
Eval doesn't make sense to me for calling chmod function. I doubt it will crash in normal situations and also you need to check if the chmod was successful anyway.

xavierba:
I just copy/paste'd the code above which does the chown. What would you suggest instead ?

racke:
Copy/paste from where?

xavierba:
From _save_binary_cache in src/lib/Conf.pm, just above the lines I've added.

ikedas:
As discussed on irc meeting, this PR is closed. #250 will survive.

Issue #244, opened 2018-03-28
sympa_soap_client: bug in logic


marcobroglia:
I've just upgraded Sympa on FreeBSD (v6.2.16 -> v6.2.24) and I experience a new error in several of my Soap tools.

The same error rise out with `sympa_soap_client.pl` script:

````
# .../sympa_soap_client.pl --soap_url http://<vh>/sympasoap --user_email=<email>
  --user_password=<pw> --service=add --service_parameters=<list>,<email>,,1
...
LOGIN....
0
        '...'

add....
Soap error :
faultstring = bug in logic. Ask developer at /usr/local/libexec/sympa/Sympa/Spindle/AuthorizeRequest.pm line 67, <DATA> line 755.

faultcode = soap:Server
````

**Comment by ikedas** [comment]
Hi @marcobroglia,

If possible, could you please apply [this patch](https://patch-diff.githubusercontent.com/raw/sympa-community/sympa/pull/251.patch) and check if the problem is solved?

**Comment by SukramD** [comment]
Hi @ikedas,
I had the same problem with sympa 6.2.30 and it seems to be fixed after I applied your patch. Thanks!

**Comment by gpearson** [comment]
I can try this as well if someone can help on what I need to do to apply the patch.

**Comment by marcobroglia** [comment]
It works. Thanks a lot.

[gpearson] The patch substitutes 3 occurrences of string `scenario_contest` with the correct `scenario_context` in file `SOAP.pm`. Then you have to restart.

**Comment by ikedas** [comment]
Hi @gpearson, thank you for offering help.
The bug at this time will be included in the next beta release (6.2.31b.1 in May). Would you please check it to confirm fixes?

**Comment by leolivier** [comment]
Hi,
I'm using 6.2.33b.2 here and I still have the same bug

**Comment by ikedas** [comment]
Hi @leolivier,
Can you show the content of system log when the error you shew occurred?

**Comment by leolivier** [comment]
Nothing very interesting, I think...
```
olivier@baloo-17 ~ $ journalctl -xe
juil. 06 12:35:11 baloo-17 kernel:  tasklet_action+0xff/0x110
juil. 06 12:35:11 baloo-17 kernel:  __do_softirq+0xf2/0x287
juil. 06 12:35:11 baloo-17 kernel:  irq_exit+0xb6/0xc0
juil. 06 12:35:11 baloo-17 kernel:  do_IRQ+0x4f/0xd0
[...]

PR #251, opened 2018-04-05
SympaSOAP crashes with add, closeList and del #244

ikedas:
This may fix issue #244.


racke:
:+1: for the fix of that somehow stupid bug.

Issue #167, opened 2018-01-17
Password has to be reset after logging out


fl02:
Sympa 6.2.24, one virtual host
Debian 9

Each time I logout of my Sympa instance I can't login again on both domains without requesting a new password. Is that a bug or known issue or could it be a configuration error on my part? Appreciate your thoughts!
Thanks!

dverdin:
Hi,
You mean that typing the password you previously defined doe not work?

crumblingwalls:
Yes exactly: I set a password, login, logout, try to login again and get the error: »ERROR () - Provided password is incorrect«. When I request and set a new password, it works again.

dverdin:
I think your password doesn't work at all. When requesting a password, you use an authentication URL . So you have a session when redefining a password and that's the session you're using until logging out.
Do you have any logs related to this problem?

crumblingwalls:
sympa.log:

> Jan 17 15:26:40 mail wwsympa[30671]: notice main:: (example.com) [robot lists.example2.com] [client 192.168.0.1] Does NOT match HTTP_HOST; setting cookie_domain to lists.example2.com
Jan 17 15:26:40 mail wwsympa[30671]: notice main:: (example.com) [robot lists.example2.com] [client 192.168.0.1] Does NOT match HTTP_HOST; setting cookie_domain to lists.example2.com
Jan 17 15:26:40 mail wwsympa[30671]: info main::do_home() [robot lists.example2.com] [session 35059202178636] [client 192.168.0.1]
Jan 17 15:26:40 mail wwsympa[30671]: info main::do_home() [robot lists.example2.com] [session 35059202178636] [client 192.168.0.1]
Jan 17 15:26:45 mail wwsympa[30671]: notice main:: (example.com) [robot lists.example2.com] [client 192.168.0.1] Does NOT match HTTP_HOST; setting cookie_domain to lists.example2.com
Jan 17 15:26:45 mail wwsympa[30671]: notice main:: (example.com) [robot lists.example2.com] [client 192.168.0.1] Does NOT match HTTP_HOST; setting cookie_domain to lists.example2.com
Jan 17 15:26:45 mail wwsympa[30671]: info main::do_login(me@example.com) [robot lists.example2.com] [session 35059202178636] [client 192.168.0.1]
Jan 17 15:26:45 mail wwsympa[30671]: info main::do_login(me@example.com) [robot lists.example2.com] [session 35059202178636] [client 192.168.0.1]
[...]

PR #261, opened 2018-04-12
Fixing issue #167 along with problem reported on the list by Pat Alle…

dverdin:
Fixing issue #167 along with problem reported on the list by Pat Allen: password was reset when subscribing to a list. When updating a user, password was systematically rehashed even if it was not a new password. Consequently, anytime a user was updated, the password was replaced by its own hash. Fixed by checking the database for a pre-existing password before computing the hash. If the hash in database and the value password present in the user data during update are the same, the password is not rehashed.


dverdin:
Killing it again. I finally understood that my fork was waaaaaay out of date, making merge impossible.
So I close this PR and will recreate it with the right repository.
Back in 5.

PR #269, opened 2018-04-16
Password spontaneous reinitialization

dverdin:
Fixing issue #167 along with problem reported on the list by Pat Allen: password was reset when subscribing to a list. When updating a user, password was systematically rehashed even if it was not a new password. Consequently, anytime a user was updated, the password was replaced by its own hash. Fixed by checking the database for a pre-existing password before computing the hash. If the hash in database and the value password present in the user data during update are the same, the password is not rehashed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## [6.2.33b.1](https://github.com/sympa-community/sympa/tree/6.2.33b.1) (2018-06-XX)
+## [6.2.33b.1](https://github.com/sympa-community/sympa/tree/6.2.33b.1) (2018-05-03)
 
 [Full Changelog](https://github.com/sympa-community/sympa/compare/6.2.32...6.2.33b.1)
 
@@ -15,6 +15,7 @@
 
 **Fixed bugs:**
 
+- Error message is missing one parameter [\#263](https://github.com/sympa-community/sympa/issues/263)
 - Erroneous typing in templates [\#266](https://github.com/sympa-community/sympa/issues/266)
 - cookie parameter protection [\#243](https://github.com/sympa-community/sympa/issues/243)
 - Spurious error on duplicate keys with admin sync [\#11](https://github.com/sympa-community/sympa/issues/11)
--- a/configure.ac
+++ b/configure.ac
@@ -25,7 +25,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 AC_PREREQ(2.60)
-AC_INIT(sympa, 6.2.32, sympa-developpers@listes.renater.fr)
+AC_INIT(sympa, 6.2.33b.1, sympa-developpers@listes.renater.fr)
 AM_INIT_AUTOMAKE([foreign -Wall -Werror 1.9 tar-pax])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 AM_PO_SUBDIRS


src/lib/Sympa/User.pm
```diff
@@ -635,23 +635,45 @@ sub update_global_user {
...
 
+    ## Check whether password is already defined.
     my $sdm = Sympa::DatabaseManager->instance;
     unless ($sdm) {
         $log->syslog('err', 'Unavailable database connection');
         return undef;
     }
 
+
+    push @sth_stack, $sth;
```
ikedas: What is use of @sth_stack?  "`my $sth;`" may suffice.

src/lib/Sympa/User.pm
```diff
@@ -635,23 +635,45 @@ sub update_global_user {
...
+        $sth = pop @sth_stack;
+        return undef;
+    }
+
+    my $current_password = $sth->fetchrow();
+
+    $sth = pop @sth_stack;
+
+    if ($values->{'password'}) {
+        if($current_password ne $values->{'password'}) {
```
ikedas: This check seems not help for me: `password_user` includes encrypted or hashed value, and it hardly match with new password in clear text form (`$values->{'password'}`).
Consequently, query above to get value of `password_user` field looks not useful.

dverdin: Actually, this test is the whole point of the PR. ;-)
The question is: what should do of a password parameter contained by $values->{'password'}?
1- if it is a new password or a clear text form of an existing password, then we can hash/encrypt it and store in the database,
2- if it is the hashed/encrypted form of an existing password we must not hash/encrypt it again because it would mean replacing the clear text form by the hashed form.
3- IT SHOULD NEVER BE the hashed/encrypted form of a new password, becsause the function assumes that hashing/encrypting must occur here, and not before.

case 1 happens when $current_password ne $values->{'password'}
case 2 happens when $current_password eq $values->{'password'}

Is it clearer this way?

ikedas: * 1 is obviously possible.
  * Is 2 possible? If it is, I think opportunistic solution you say should not be taken. Solution should be that: We correct  the code so that any callers of update_global_user() won't hash/encrypt `$values->{password}`.
  * If 2 is solved, 3 does not matter: Even if user submitted a text looks like hashed/encrypted, it will be treated as legitimate password, and it is encrypted/hashed and stored in table.

----
BTW the way of processing also looks problematic: The way "fetching a row, checking fetched value, then updating the row" is not atomic.

dverdin: Oh yes, 2 is possible: anytime we update an existing user object by passing the sub the content of said user.
That's the reason why the bug was reported after all. That is exactly what happened.
I think that checking data before updating them is not such  a bad idea. And password is the only case in which we have to make such check.

ikedas: I realized that, in the first place, the assumption that password can be updated along with other user attributes is inappropriate.  In fact, there are no cases updating password and other fields simultaneously.  Proper solution may be either:
  - Check content of argument `$value` to construct query updating only available field(s) (cf. `Sympa::List::update_list_member()` and `Sympa::List::update_list_admin()`.
  - Modify `update_global_user()` not to change password field, and add new function to change password, such as `update_user_password()`.

racke:
@dverdin This PR isn't usable anymore, because it contains many unrelated commits and the remote source has gone too. Please resubmit (e.g. by cherry-picking the necessary commits in a local branch).

racke:
@dverdin agreed to resubmit.

PR #320, opened 2018-05-23
Add copyright notice to sympa.js

xavierba:
sympa.js is missing a copyright notice.
I traced back the code to 2010 : https://github.com/sympa-community/sympa/commits/8af6368fe8acec2caa0c66298e6614eed969d6b3/src/etc/script/js/sympa.js
@dverdin and @ikedas are the ones who significantly touched the code.
I chose to use a short copyright notice rather than the full GPL boilerplate in order to not add too much to the transfer size of the javascript to the client.
@dverdin , @ikedas,  do you both agree to the licensing (same as sympa itself) and the shape of the copyright notice ?

### DIFF
--- a/www/js/sympa.js
+++ b/www/js/sympa.js
@@ -1,4 +1,7 @@
-/* $Id$ */
+/*
+ * Copyright 2010-2018 The Sympa Community. Licensed under GNU GPL v2
+ * See license text at https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
+ */
 
 // To confirm on a link (A HREF)
 function refresh_mom_and_die() {


dverdin:
Agreed. I don't have a strong opinion on license anyway. Getting the same everywhere looks good to me. So yes, please, go ahead!

ikedas:
I have no objection to adding notices about license (it is recommended/required by GPL).
I think we'd be better to adopt single long format, not only because of its strictness but also less controversy (most of us are not law professionals and discussion will not be so productive).

I think size does not matter in this case.
- As the script will be cached, effect by size is practically ignorable.
- At the first place, current web interface operates slightly slow and users might not notice to addition of traffic by 1 kB. :stuck_out_tongue_winking_eye:

In the future, we may strip long comments by script aggregator (minifier) before sending them so that we can follow faster frontend.

ikedas:
I think context should be considered.

That proposal concerns the case that js sources themelves would be distributed freely, and they would be discriminated by LibreJS appropriately.  Also, it assumes use of [Magnet links](https://en.wikipedia.org/wiki/Magnet_URI_scheme) to identify js sources.

It looks not applicable to js sources in general. In fact, js sources formiing parts of LibreJS itself do not use this notation (see [repo](http://git.savannah.gnu.org/cgit/librejs.git/tree/)).

----
Since this issue may not need early conclusion, I'll leave discussion and will check progress at times.

xavierba:
Here's the long version of the copyright notice: https://github.com/xavierba/sympa/commit/26e95cf44332a2a7f42262e6c6b4092f1402e0a0

Either short (https://github.com/xavierba/sympa/commit/23b5d411859eb5c71c09f9514135f552d3b8a366) or long (https://github.com/xavierba/sympa/commit/26e95cf44332a2a7f42262e6c6b4092f1402e0a0) copyright notice can be used once #369 has landed (all code in sympa.js will truly be sympa code then).

My understanding is the long version should be preferred.

ikedas:
I merged long version by xavierba. This PR is closed for the present.
Thanks for input!

Issue #175, opened 2018-01-30
Both send.confidential and send.private scenari files use the same gettext text


qosobrin:
The title is self explanatory. This error translates into users having, in the dropdown menu, two entries that are equal, so they can not differenciate which is the one they want to pick.

ikedas:
Fixed.  Thanks for reporting bug!

PR #328, opened 2018-05-27
[-bug] Title of send.confidential and send.private are the same.

ikedas:
This may fix issue #175.


racke:
I think we can improve the text "messages from the others are discarded".

Either:
"messages from others" (you don't use _the_ in this context)
"messages from everyone else"

Of course, a native speaker might be able to provide an even better one ..

ikedas:
@racke, thanks! I trust your english skill.  I'll remove _the_ and push again.

PR #330, opened 2018-05-29
WWSympa: wwsympa_url would be optional

ikedas:
As of current version (6.2.32), mandatory parameters in `sympa.conf` are:
- `domain`
- `listmaster`
- `wwsympa_url`

The last would be optional to run services without web interface.

This PR makes `wwsympa_url` parameter optional, and fixes some mail templates to manage situation without web interface.


ikedas:
Update.

Change:
- Administratora have to add `wwsympa_url` parameter to `robot.conf` (if `robot.conf` is there) explicitly if corresponding domain will have web interface.  Previously `http://`_domain_`/sympa` was used in `robot.conf` by default.

PR #352, opened 2018-06-23
Remove tabs in default tt2 templates + some indentation changes

ldidry:
I recently updated my Sympa and thus dived in the templates to update my changes. I saw a lot of tab indentation mixed with spaces indentation, which make templates hard to read if your tab width setting is not the one used by the writer of the template (and the tab width is not always the same accross the different templates).

So I removed all tabs and replaced them by spaces. I also made indentation improvments in some templates (not all: it would be a huge work, I may do it in a second time).

If you do a `git diff --ignore-all-space` on my commit, you'll see that I didn't touch much real code (I replaced some tags that were in caps, small things like that).

### DIFF
--- a/default/mhonarc-ressources.tt2
+++ b/default/mhonarc-ressources.tt2
@@ -20,11 +20,11 @@ CALENDAR
         <div>
         <ul class="calendar">
         <li class="calendarLinksInactive">($tag$% year.key %$tag$)</li>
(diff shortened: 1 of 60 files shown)


xavierba:
The non-whitespace changes are indeed very few and look sane. Should be good to merge.

PR #357, opened 2018-06-26
Use rsa-sha256 for DKIM signatures

FabianHenneke:
Sympa currently uses rsa-sha1 for DKIM signature attached to outgoing mails. It [has been shown](https://shattered.io/) that SHA-1 can no longer be considered resistant to collisions in practice, which means that it is inadequate to be used with DKIM. Since Mail::DKIM supports rsa-sha256, this is easy to fix.

### DIFF
--- a/src/lib/Sympa/Message.pm
+++ b/src/lib/Sympa/Message.pm
@@ -493,7 +493,7 @@ sub dkim_sign {
     }
     # create a signer object
     my $dkim = Mail::DKIM::Signer->new(
-        Algorithm => "rsa-sha1",
+        Algorithm => "rsa-sha256",
         Method    => "relaxed",
         Domain    => $dkim_d,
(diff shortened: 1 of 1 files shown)


racke:
Makes sense to me :+1: Do you know since which version Mail::DKIM supports rsa-sha256?

FabianHenneke:
Since version 0.17 from 2006, so this should not pose any restrictions.

racke:
Indeed :innocent: , thanks for the report!

xavierba:
While this is not a security issue in sympa code by itself, wouldn't that still make sense this PR be tagged "security" too ?

FabianHenneke:
I agree with this assessment. As sha1 has been broken in practice, this is
not just a theoretical issue.

ikedas:
@xavierba ++

Issue #362, opened 2018-07-03
sceneri that check subscriber/editor/owner/listmaster should require auth


taggart:
Many of the existing sceneri do checks for is_subscriber/is_editor/is_owner/is_listmaster and blindly trust smtp requests, which can be forged. One can generate a list of potential ones with:

`default/scenari$ grep is_ * |grep smtp |grep do_it`

(these are always wrong, but in some cases an earlier line might be firing and preventing from getting here, still it should be fixed).

It's far too easy to forge email and any assumptions made in the past about a sympa server only being for private use are no longer valid, even private servers have to be secure against attackers these days. So it's time for the unauthenticated sceneri to go away.

In the past, many sceneri had both a non-authenticated version and an authenticated version, currently there are 19 different ones with 'auth' in the name. And 'auth' in the name usually meant it did a request_auth, although in some cases dkim or smime was also acceptable to establish a legit request.

One thing complicating the problem is that it's difficult to rename or remove scenari because once you have existing lists pointing to a particular name in their config you will need to migrate them to something else before making changes. However it's possible to change the content of scenari and leave the name the same. I think it might be possible to update all the non-auth versions of sceneri to use auth (probably resulting in them being identical to the auth versions). Then maybe the sympa upgrade script could make one a symlink, or migrate existing lists away from one name or the other. New installs could get a smaller set of sceneri.

Version
-------
6.2.32 and older.

Installation method
-------------------

Expected behavior
-----------------

Actual behavior
---------------

Additional information
----------------------

ikedas:
IMO `dkim` authentication method is not so stricter than `smtp` method.  Because DKIM authenticates just a sender domain, not an identity of each sender.

My suggestion is that `smtp` and `dkim` authentication methods would be merged, and that new configuration switch to force DKIM check would be introduced (if this option is enabled and DKIM check fails, `[sender]` would be `nobody`).

Issue #179, opened 2018-02-02
Edit List Config menu items appear ordered by scenario name instead of by gettext description


qosobrin:
When trying to define the properties of a list in "Edit List Config" you have to face many drop down menus that have their options ordered by the scenario name instead of by the appropriate gettext description. This makes very odd, if not difficult, to figure out the best option to use.

For example, if you try to set the "send" option in a list you are faced with this list:

closed (closed)
restricted to subscribers (confidential)
restricted to subscribers (default)
Moderated, no authentication needed if DKIM signature from editor is OK (editordkim)
Moderated (editorkey)
Moderated, even for moderators (editorkeyonly)
Moderated, need  authentication  from editor (editorkeyonlyauth)
restricted to local domain (intranet)
restricted to local domain and subscribers (intranetorprivate)
Newsletter, restricted to moderators (newsletter)
Newsletter, restricted to moderators after confirmation (newsletterkeyonly)
restricted to subscribers (private)
Moderated, restricted to subscribers (privateandeditorkey)
Moderated, for non subscribers sending multipart messages (privateandnomultipartoreditorkey)
restricted to subscribers with previous md5 authentication (privatekey)
Moderated, for subscribers and even moderators themself (privatekeyandeditorkeyonly)
Private, moderated for non subscribers (privateoreditorkey)
Private, confirmation for non subscribers (privateorpublickey)
restricted to subscribers and checked smime signature (private_smime)
public list (public)
anyone no authentication if DKIM signature is OK (publickey)
public list multipart/mixed messages are forwarded to moderator (publicnoattachment)
public list, Bcc rejected (anti-spam) (public_nobcc)
public list multipart messages are rejected (publicnomultipart)

As you can see, several related or similar options are scattered across the whole list. A more useful approach would be to have the options sorted by their gettext description. The same list would show like this:

anyone no authentication if DKIM signature is OK (publickey)
closed (closed)
Moderated (editorkey)
Moderated, even for moderators (editorkeyonly)
Moderated, for non subscribers sending multipart messages (privateandnomultipartoreditorkey)
Moderated, for subscribers and even moderators themself (privatekeyandeditorkeyonly)
Moderated, need  authentication  from editor (editorkeyonlyauth)
Moderated, no authentication needed if DKIM signature from editor is OK (editordkim)
Moderated, restricted to subscribers (privateandeditorkey)
Newsletter, restricted to moderators (newsletter)
[...]

PR #365, opened 2018-07-03
Sceneri title changes

taggart:
Here is a collection of scenari related stuff, mostly title.gettext changes to make things more consistent. There are a couple changes that could be considered related to #362 where an authenticated version of a type didn't exist. In those cases the patch adds one, but maybe it would be better to add the authentication to the existing non-auth version?

Feel free to just cherry pick whatever makes sense and also let me know if you'd prefer that I break things up into multiple PRs.

### DIFF
--- a/default/scenari/info.conceal
+++ b/default/scenari/info.conceal
@@ -1,4 +1,4 @@
-title.gettext restricted to subscribers - Silent rejection otherwise.
+title.gettext visible only to subscribers, silent rejection otherwise.
 
 is_subscriber([listname],[sender])   	smtp,md5,smime -> do_it
 is_listmaster([sender])			smtp,md5,smime -> do_it
--- a/default/scenari/info.open
+++ b/default/scenari/info.open
@@ -1,3 +1,3 @@
-title.gettext for anyone
+title.gettext publically visible
 
 true()   smtp,md5,smime -> do_it
--- a/default/scenari/info.private
+++ b/default/scenari/info.private
@@ -1,4 +1,4 @@
-title.gettext restricted to subscribers
+title.gettext visible only to subscribers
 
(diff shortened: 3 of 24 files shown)


ikedas:
@taggart, I feel we would be better to divide works: Adding auth versions (suggested by #362), and correcting titles (currently proposed here).

One idea: Instead of submitting single PR, how about opening another issue page to discuss about titles?

qosobrin:
Why don't you have https://github.com/sympa-community/sympa/issues/179 into account? In my institution we ended by renaming scenari files with numeral prefixes. For example, the add menu now shows these scenarios in this order:

```
add.auth            restricted to owner with authentication
add.authdkim        restricted to owner without authentication if DKIM signature is OK.
add.closed          add impossible
add.owner           add performed by list owner does not need authentication
add.owner_notify    add performed by owner does not need authentication (notification)
add.ownerdkim       add performed by list owner does not need authentication if DKIM signature OK
```

But if you rename the scenari files from public to restricted you will get something like:

```
add.1-owner         add performed by list owner does not need authentication
add.2-owner_notify  add performed by owner does not need authentication (notification)
add.3-ownerdkim     add performed by list owner does not need authentication if DKIM signature OK
add.4-authdkim      restricted to owner without authentication if DKIM signature is OK.
add.5-auth          restricted to owner with authentication
add.6-closed        add impossible
```

This makes life easier for users to select the correct scenario since now they know the scenarios proposed in the drop-down menus have a logical order. And furthermore, an exercise like this will allow you to identify scenarios that are useless or are equal (for example, `add.authdkim` is basically the same as `add.ownerdkim` and one of them could be removed). 

Best regards,

ikedas:
@qosobrin , I think @taqggart does not propose about ordering of the scenarios.

Issue #372, opened 2018-07-11
Error on closing a list that was already closed from command line 


xavierba:
Version
-------
6.2.34

Installation method
-------------------
rpm

Expected behavior
-----------------
$sudo sympa.pl --close_list=test2
/var/lib/sympa/sympa_aliases: 5 aliases, longest 58 bytes, 316 bytes total
/etc/aliases: 84 aliases, longest 16 bytes, 890 bytes total
/etc/sympa/aliases.sympa.sendmail: 6 aliases, longest 53 bytes, 293 bytes total
close_list [notice] List has been closed
$ sudo sympa.pl --close_list=test2
close_list [error] List is already closed

Or whatever output, but not an error...

Actual behavior
---------------
$sudo sympa.pl --close_list=test2
/var/lib/sympa/sympa_aliases: 5 aliases, longest 58 bytes, 316 bytes total
/etc/aliases: 84 aliases, longest 16 bytes, 890 bytes total
/etc/sympa/aliases.sympa.sendmail: 6 aliases, longest 53 bytes, 293 bytes total
close_list [notice] List has been closed
$ sudo sympa.pl --close_list=test2
Can't use string ("user") as a HASH ref while "strict refs" in use at /usr/sbin/sympa.pl line 1079.

Additional information
----------
The sympa log correctly reports that the list was already closed:
Jul 11 15:34:12 mail sympa[19917]: info main::_load() Configuration file read, default log level 0
Jul 11 15:34:12 mail sympa[19917]: notice main:: Sympa 6.2.34 Started
Jul 11 15:34:12 mail sympa[19917]: notice Sympa::Spindle::ProcessRequest::_twist() Processing Sympa::Request <action=close_list;current_list=test2@domain.tld>
Jul 11 15:34:12 mail sympa[19917]: err main::#543 > Sympa::Spindle::spin#95 > Sympa::Request::Handler::close_list::_twist#67 List Sympa::List <test2@domain.tld> is already closed: cannot close it again

ikedas:
@xavierba, can you check this PR?

xavierba:
@ikedas, I can confirm #373 fixes #372, thank you.

[...]

PR #373, opened 2018-07-12
close_list may crash with typos.

ikedas:
This may fix issue #372.


xavierba:
I can confirm this fixes #372, thank you.

PR #377, opened 2018-07-13
Improve performance of purge operation

cgx:
## Description

When purging a closed list, all tasks are instantiated which can be a very long process when you have thousands of lists defined.

## Impacts

This PR improves the purge operation by only instantiating tasks relevant to the list being purged.


ikedas:
Merged. Thanks for improvement!

PR #394, opened 2018-08-14
Refactoring task_manager.pl

ikedas:
- Introducing Sympa::Spool::Task & Sympa::Spindle::ProcessTask.
- [change] Move task files to `tasks/` subdirectory. Previously `list_task_models/` and `global_task_models/` directories were used. Task files in older directories will be automatically copied to new places during upgrading process.
- Removed unused task files from distribution: `expire.*` (see also note), `crl_update.*` and `chk_cert_expiration.*` which have not been implemented yet.
- And some related fixups.

----
Additional note:
  - As [requested](https://listes.renater.fr/sympa/arcsearch_id/sympa-users/2018-05/cbcdaeb6-cb61-5df2-4b81-10a825e7b643%40dcs.gla.ac.uk), `expire.*` task models would be better to be restored (reimplemented). I'm planning to submit additional PR after merger of this PR.


ikedas:
This PR depends on PR #377 & #388.

aruhri:
It would be nice to have tasks in Sympa. Maybe not only predefined ones. Especially for expiry/GDPR-acceptance. But I can imagine more scenarios where tasks could be helpful.

PR #398, opened 2018-08-22
Avoid rehashing user password hashes in {add,update}_global_user()

mpkut:
In 6.2.32 and 6.2.34, our testing shows that using the GUI to add an existing address to a list causes the user's current password hash to be rehashed and stored. This renders the user's password unusable and effectively forces a password reset sequence. The behavior seems to correspond with the spontaneous password reinitialization problem discussed in #269. 

This targeted fix uses `hash_type()` to detect password hashes and avoid rehashing them. For the sake of consistent behavior, both `update_global_user()` and `add_global_user()` are updated.

The change does not address all of the design concerns of #269 but in our testing it does prevent the seemingly spontaneous password resets caused by being added to a list using the GUI.

### DIFF
--- a/src/lib/Sympa/User.pm
+++ b/src/lib/Sympa/User.pm
@@ -642,9 +642,17 @@ sub update_global_user {
 
     ## use hash fingerprint to store password
     ## hashes that use salts will randomly generate one
-    $values->{'password'} =
-        Sympa::User::password_fingerprint($values->{'password'}, undef)
-        if ($values->{'password'});
+    ## avoid rehashing passwords that are already hash strings
+    if ($values->{'password'}) {
+        if (defined(hash_type($values->{'password'}))) {
+            $log->syslog('debug',
+                         'password is in %s format, not rehashing',
+                         hash_type($values->{'password'}));
+        } else {
+            $values->{'password'} =
+                Sympa::User::password_fingerprint($values->{'password'}, undef);
+        }
+    }
 
     ## Canonicalize lang if possible.
     $values->{'lang'} = Sympa::Language::canonic_lang($values->{'lang'})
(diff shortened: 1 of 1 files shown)


ikedas:
This looks a workaround for issue #167.  We would be better to add it to the next beta (6.2.35b.1) and to see what will occur.  So I'll merge.

ikedas:
I made additional changes to track the bug: https://github.com/sympa-community/sympa/commit/e02945f810e0c4f01511d613bbd0a5d48a2d35fe

mpkut:
Thank you!

BTW I would not see this as a workaround for #167; that seems to be a password reset loop. We have seen those too, in two varieties: 1) stored form data in browsers, and 2) people remaining on the password reset page even when they have successfully chosen a password. In both cases, clearing cookies and manually entering the form data seems to help.

In contrast, the band-aid in this PR is about keeping other people from invalidating your password hash simply by adding you to a list using WWSympa. Of course, encountering that situation will increase your chance of entering one of the password reset loop scenarios, which is what we've seen lately.

ikedas:
Hi @mpkut,

Honestly, I have not grasped what causes the bug reported in #167: Password field is overwritten unexpectedly.  I hope your first aid will help making it clear.  When sympa spews err log, please let us know.

mpkut:
Hello @ikedas:We installed the updated logging in production tonight and immediately started seeing these messages. Examples: 
Aug 29 20:50:13 utlists-sy-p01 wwsympa[6253]: err main::#1573 > main::do_add#7015 > Sympa::Spindle::spin#95 > Sympa::Request::Handler::add::_twist#135 > Sympa::User::save#203 > Sympa::User::update_global_user#641 Bug in logic: Password is in bcrypt format, not rehashing

mpkut:
Hello @ikedas 
We installed the updated logging code this evening and pretty much immediately started seeing log messages. Here is a brief sample. 

[rehash-errors.txt](https://github.com/sympa-community/sympa/files/2334352/rehash-errors.txt)

PR #410, opened 2018-09-05
In Sub _merge_msg Content-Disposition: attachement in multipart-messa…

pveith:
This should resolve Issue 409 as far as i tested it (Invitations and attachment files with special characters). Further Information see there.

### DIFF
--- a/src/lib/Sympa/Message.pm
+++ b/src/lib/Sympa/Message.pm
@@ -1264,7 +1264,7 @@ sub _merge_msg {
 
     # Check for attchment-part, which should not be changed
     my $cdisposition = $entity->head->mime_attr('Content-Disposition');
-    if ($cdisposition and $cdisposition == 'attachment') {
+    if ($cdisposition and lc($cdisposition) eq 'attachment') {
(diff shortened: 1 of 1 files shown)


src/lib/Sympa/Message.pm
```diff
@@ -1262,6 +1262,13 @@ sub _merge_msg {
         return $entity;
     }
 
+    # Check for attchment-part, which should not be changed
+    my $cdisposition = $entity->head->mime_attr('Content-Disposition');
+    if ($cdisposition and $cdisposition == 'attachment') {
```
ikedas: eq should be used.

ikedas: and both upper/lowercase values should be considered.

pveith:
I installed the PR on our production sympa server, which now emits quite a few notices in the log. Everything works as expected though, as far as i can see; i would be inclined to drop the log notice.

ikedas:
@pveith, if you haven't experienced problem, I'll merge this pr.

pveith:
No Problems detected this week. After using it in our production server for a week, i think the PR works just as expected.

ikedas:
Merged. Thanks!

PR #416, opened 2018-09-14
Drop support for htmlArea.

ikedas:
It was gone years ago.


racke:
Looks good to me :+1:

Issue #153, opened 2018-01-05
Adding ARC support


ikedas:
Current beta: Sympa 6.2.37b.2

### ToDo

- [x] Making changes on Sympa
- [x] Documentation
- [x] Beta testing

### Documentation (draft)

- Sympa Administration Manual: [DKIM and ARC features for Sympa](https://sympa-community.github.io/manual/customize/dkim-arc.html)

### References

Related posts:
  - [\[sympa-developpers\] Adding ARC supprt to Sympa](https://listes.renater.fr/sympa/arcsearch_id/sympa-developpers/2018-01/alpine.OSX.2.21.1801042045010.25310%40ary.qy)

Specification:
  - [Authenticated Received Chain Overview](https://dmarc.org/presentations/ARC-Overview-2016Q3-v01.pdf) by DMARC.org
  - [Authenticated Received Chain (ARC) Protocol](https://tools.ietf.org/html/draft-ietf-dmarc-arc-protocol-10), draft-ietf-dmarc-arc-protocol
  - [Recommended Usage of the Authenticated Received Chain (ARC)](https://tools.ietf.org/html/draft-ietf-dmarc-arc-usage), draft-ietf-dmarc-arc-usage

jrlevine:
General flow:
* as message is received, see if there is an Authentication-Results header. If there is, do ARC validation on incoming message and edit arc=<whatever> result into the A-R header or arc=none if none
* as outgoing message is signed, look for that A-R header again, use the incoming ARC result in ARC signing which can happen at the same time as DKIM signing
* add ARC headers above DKIM header(s) on the way out
I can provide sample code if it'd be useful

ikedas:
@jrlevine, thanks for guidance. Please do provide sample!

jrlevine:
[prearc.txt](https://github.com/sympa-community/sympa/files/1629547/prearc.txt)
Here's the prepass I use to add the incoming arc-chain to the Authentication-Results header. I run it on deliver as:
|bin/prearc | bin/queue <listname>

jrlevine:
Here's a simplified version of my post-pass ARC and DKIM signer. (I used to use mj2 and it was easier to keep using this than to integrate it into Sympa.) You can see how the ARC signature is added in parallel with DKIM signature(s).
[sympamail.txt](https://github.com/sympa-community/sympa/files/1629551/sympamail.txt)

dverdin:
[...]

PR #419, opened 2018-09-22
Arc signatures

jrlevine:
Patches to support ARC seals in addition to DKIM signatures

Notes:

- ARC config is similar to DKIM config. If you have DKIM working, and enable arc_feature in robots.txt and set arc_srvid to the srvid in your MTA's Authentication-Results, it should just work
- ARC uses the same kind of keys as DKIM and defaults to the same domain, selector, and keyfile as DKIM, which is usually correct
- ARC will only sign a message if Sympa also adds a DKIM signature. Set Sympa to DKIM sign everything, since it can only help deliverability
- ARC needs an Authentication-Results header applied by the incoming MTA. If your MTA doesn't already do that, there are drop-in milters
- ARC only applies to forwarded messages, so there are no ARC seals on  Sympa generated mail such as welcome messages and digests (but do DKIM sign them)


ikedas:
Related to #153.

src/lib/Sympa/ListDef.pm
```diff
@@ -2153,8 +2153,75 @@ our %pinfo = (
...
+                    'The ARC "d=" tag, is the domain of the sealing entity. The list domain MUST be included in the "d=" domain',
+                'format'     => '\S+',
+                'occurrence' => '0-1',
+                'default'    => {'conf' => 'arc_signer_domain'}
+            },
+        },
+        'occurrence' => '0-1'
+    },
+
+    'dkim_signature_apply_on' => {
```
ikedas: dkim_signature_apply_on is duplicated.

default/edit_list.conf
```diff
@@ -140,6 +140,9 @@ dkim_parameters			owner,privileged_owner 		hidden
 dkim_signature_apply_on		listmaster	 		write
 dkim_signature_apply_on		owner,privileged_owner 		hidden
 
+arc_parameters			listmaster			write
+arc_parameters			owner,privileged_owner		hidden
+
```
ikedas: arc_feature would be better to be added.

src/lib/Sympa/ListDef.pm
jrlevine: Correct, cut and paste error. Just delete the second one.

default/edit_list.conf
jrlevine: Yes, added

ikedas:
I found no particluar problems anymore with review on desktop. Let's merge this pr and look it works fine.

PR #421, opened 2018-09-24
Create ${expldir}/${robot} directory if it does not exists

k0lter:
Sometimes (while manually creating a robot) in testing or production mode, it could be useful to have this feature implemented.

1. It makes possible to load the robot configuration (from the wwsympa side) without issues even if ${expldir}/${robot} does not exists
2. It makes possible to create a first list even if ${expldir}/${robot} does not exists

Feel free to reject it if you think it is not suitable.

Maybe this PR could help people to save time.

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -842,6 +842,9 @@ sub load {
 
     ## Set of initializations ; only performed when the config is first loaded
     if ($options->{'first_access'}) {
+        if ($robot && (! -d "$Conf::Conf{'home'}/$robot")) {
+            mkdir "$Conf::Conf{'home'}/$robot", 0775
+        }
         if ($robot && (-d "$Conf::Conf{'home'}/$robot")) {
             $self->{'dir'} = "$Conf::Conf{'home'}/$robot/$name";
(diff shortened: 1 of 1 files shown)


ikedas:
Hi @k0lter,

At the moment we have to support old-style list path `${expldir}/${listname}/` along with `${expldir}/${robot}/${listname}/` (cf. [doc](https://sympa-community.github.io/manual/customize/basics-configuration.html#configuration-files): See Note).

We may have to give up old-style path, and `${robot}` directory will be created at the time of list creation.

src/lib/Sympa/List.pm
```diff
@@ -842,6 +842,9 @@ sub load {
 
     ## Set of initializations ; only performed when the config is first loaded
     if ($options->{'first_access'}) {
+        if ($robot && (! -d "$Conf::Conf{'home'}/$robot")) {
+            mkdir "$Conf::Conf{'home'}/$robot", 0775
```
ikedas: To keep compatibility with older style path `${expldir}/${listname}`, I think this is better:
```
        if ($robot && (! -d "$Conf::Conf{'home'}/$robot")
            and $robot ne $Conf::Conf{'domain'}) {
            mkdir ...
```
in case some lists have already had `${expldir}/${name}` directories.

PR #423, opened 2018-09-28
split the test in separated ones

eiro:
the previous message was unclear: you didn't know if it failed to create
the directory or to apply the ownership.

i split the test to be able to write a more informative message.


racke:
Yes, but in my opinion it would be more useful to have a create_dir method in Sympa::Tools::File which does both things and returns an useful error message.

eiro:
ack ... i change this and come back

racke:
Please do and choose a less confusing subject next time.

PR #425, opened 2018-09-30
add a Sympa::Tools::File::create_dir

eiro:
related to #423. @racke asked me to isolate the code in a function. we didn't talk about the parameters so  i just take the dir to create.

possible evolutions of this pr: 
* remove the possible usage of $_ 
* use a signature identical to set_rights

### DIFF
--- a/src/lib/Sympa/Tools/File.pm
+++ b/src/lib/Sympa/Tools/File.pm
@@ -98,6 +98,29 @@ sub del_dir {
     }
 }
 
+sub create_dir {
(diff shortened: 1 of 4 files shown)


ikedas:
@eiro,

I’m very sorry.  Here is a place to submit improvements or bug fixes of a particular software, Sympa.

When you add a function, it should either replace another existing function, or add new feature to Sympa.  Please don’t add a function never used anywhere.

Closed.

PR #426, opened 2018-09-30
add create_dir and use it in Bulk to make error messages more useful

eiro:
when a bulk can't make its tree, there was no way to know if the
directory wasn't created or it it failed to get the good rights.

create_dir have error messages for both cases. Bulk.pm now benefit
of it.

related to #423 and #425

### DIFF
--- a/src/lib/Sympa/Bulk.pm
+++ b/src/lib/Sympa/Bulk.pm
@@ -74,25 +74,17 @@ sub _create_spool {
     my $self = shift;
 
     my $umask = umask oct $Conf::Conf{'umask'};
-    foreach my $directory (
-        $Conf::Conf{queuebulk},     $self->{msg_directory},
-        $self->{pct_directory},     $self->{bad_directory},
-        $self->{bad_msg_directory}, $self->{bad_pct_directory}
-    ) {
-        unless (-d $directory) {
-            $log->syslog('info', 'Creating spool %s', $directory);
-            unless (
-                mkdir($directory, 0755)
-                and Sympa::Tools::File::set_file_rights(
-                    file  => $directory,
-                    user  => Sympa::Constants::USER(),
-                    group => Sympa::Constants::GROUP()
-                )
-            ) {
-                die sprintf 'Cannot create %s: %s', $directory, $ERRNO;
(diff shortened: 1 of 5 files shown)


ikedas:
@eiro,
Please stop submitting pull requests and issues for a while.  And please ask your colleagues why I said so.

eiro:
@ikedas, please tell me clearly how to achieve this PR

racke:
The pull request contains _unnecessary_ and _unrelated_ changes. Submitting the smallest possible set of changes is good practice and makes the review lot easier.

ikedas:
> The pull request contains _unnecessary_ and _unrelated_ changes. Submitting the smallest possible set of changes is good practice and makes the review lot easier. 

I agree.  However, I want to point out that not containing _unnecessary_ things doesn't mean not _fully_ containing necessary things.  That is, in this pr, the new function create_dir() replaces only one place of code, even if the same or similar logic can be found in more than one places in the sources.

At the beginning, it is no use testing a function used only once.  If the submitter aimed to extend or modify behavior of a API, they might modify single place.  However, if they aimed to create a new API, they should review entire sources to modify all related things.  I think lack of such consideration made this pr a set of _unrelated_ things.

May this pr be closed?

racke:
> > The pull request contains _unnecessary_ and _unrelated_ changes. Submitting the smallest possible set of changes is good practice and makes the review lot easier.
> 
> I agree. However, I want to point out that not containing _unnecessary_ things doesn't mean not _fully_ containing necessary things. That is, in this pr, the new function create_dir() replaces only one place of code, even if the same or similar logic can be found in more than one places in the sources.
> 
> At the beginning, it is no use testing a function used only once. If the submitter aimed to extend or modify behavior of existing API, they might modify single place. However, if they aimed to create a new API, they should review entire sources to modify all related things. I think lack of such consideration made this pr a set of _unrelated_ things.
> 
> May this pr be closed?

I would argue about review entire sources, it could have served as Proof of Concept if done correctly and expanded to other parts of the source code.

Anyway, I concur with closing the PR ... @eiro didn't fix the concerns and he can open a better one if he wants to.

PR #441, opened 2018-10-13
first attempt to a complete, documented test suite for bulks

eiro:
taking care of the previous review testing creation, store, remove and quarantine methods of a bulk, this is just the test suite without code improvement (delayed for next PR)

### DIFF
--- a/t/bulk.t
+++ b/t/bulk.t
@@ -2,7 +2,6 @@
 
 use strict;
 use warnings;
-use FindBin qw( $Bin );
 use lib qw( t/lib );
 use Test::More;
 use Sympa::Tools::File;
@@ -30,8 +29,8 @@ note 'also: patches are welcome';
     : sub {
         shift; # $self isn't expected anymore
         my ( $level, $fmt, @params ) = @_;
-        note sprintf "syslog $level: $fmt", @params; }
-    ;
+        note sprintf "syslog $level: $fmt", @params;
+    }
 }
 
 # get default values to use from Sympa::ConfDef
@@ -90,9 +89,7 @@ sub get_definitions_of {
 # i (eiro) don't know if it's a bug or a feature
 
 use Cwd;
-$Conf::Conf{queuebulk} =
-    Cwd::abs_path
-    Sympa::Constants::SPOOLDIR."/bulk";
+$Conf::Conf{queuebulk} = Cwd::abs_path Sympa::Constants::SPOOLDIR . "/bulk";
 
 # we need to be sure that new files are created by the code below
 # so we delete the old queuebulk
@@ -104,12 +101,10 @@ for ( $Conf::Conf{queuebulk} ) {
 }
 
 # create a new Sympa::Message from the content of t/samples/unsigned.eml
-my %sample;
-@sample{qw( file msg )} =
-    map +(
-        $_,
-        Sympa::Message->new_from_file($_)
-    ), "t/samples/unsigned.eml";
+my %sample = (
+    file => "t/samples/unsigned.eml",
+    msg  => Sympa::Message->new_from_file("t/samples/unsigned.eml")
+);
 
 # store a message and returns a hashref with its
 # qw( total_packets marshalled ) as keys.
@@ -139,7 +134,7 @@ sub ok_no_next_from_empty_bulk {
     my ( $msg, $file ) = $bulk->next;
     ok +( not defined $file )
         => "no next message $desc"
-            or note "next: $file";
+            or diag "next: $file";
 }
 
 sub ok_next_message {
@@ -200,8 +195,9 @@ for my $context ('fresh queuebulk', 'existing queuebulk') {
 # because those messages are just kept in quarantine, they aren't removed
 # so they should remain on the quarantine directory.
 
-$_ = Sympa::Bulk->new->{bad_msg_directory};
-ok +( 2 == map $_, <$_/*> )
+my @remaining_files = glob Sympa::Bulk->new->{bad_msg_directory} . "/*";
+
+ok +( 2 == @remaining_files )
     => "2 messages in quarantine";
 
 done_testing;
--- a/t/bulk.t
+++ b/t/bulk.t
@@ -1,0 +1,167 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use FindBin qw( $Bin );
+use lib qw( t/lib );
+use Test::More;
+use Sympa::Tools::File;
+use Sympa::Message;
+use ok 'Sympa::Bulk';
+use English;
+
+# during a test, syslog would be helpful as a note
+# so *Sympa::Log::syslog is diverted
+#
+# by default, syslog remains silent
+
+{ no warnings;
+    *Sympa::Log::syslog = 1 #  $ENV{SYMPA_TEST_SILENT_SYSLOG}
+    ? sub {}
+    : sub {
+        shift; # $self isn't expected anymore
+        my ( $level, $fmt, @params ) = @_;
+        note sprintf "syslog $level: $fmt", @params; }
+    ;
+}
+
+# this is the basic (ideally the only one keys required)
+# config to make this test work
+
+%Conf::Conf = (
+    # Sympa::Bulk::store needs
+    sympa_packet_priority => 5,
+    sympa_priority        => 1,
+    email                 => 'sympa',
+    domain                => 'lists.example.com',
+    sender_headers        => 'Resent-From,From,Return-Path',
+    list_check_suffixes   => 'list_check_suffixes',
+    umask                 => '027',
+);
+
+# we need an absolute path there because in Sympa/Bulk.pm:109
+#
+#    $self->{_metadatas} = [
+#        sort grep {
+#                    !/,lock/
+#                and !m{(?:\A|/)(?:\.|T\.|BAD-)}
+#                and -f ($self->{pct_directory} . '/' . $_)
+#        } glob $self->{_glob_pattern}
+#    ];
+#
+# as we are already in $self->{pct_directory}, -f will always fail with a
+# relative path.
+#
+# i (eiro) don't know if it's a bug or a feature
+
+use Cwd;
+$Conf::Conf{queuebulk} =
+    Cwd::abs_path
+    Sympa::Constants::SPOOLDIR."/bulk";
+
+# we need to be sure that new files are created by the code below
+# so we delete the old queuebulk
+
+for ( $Conf::Conf{queuebulk} ) {
+    Sympa::Tools::File::del_dir $_;
+    ok +( ! -e ) => "remove queuebulk";
+}
+
+# create a new Sympa::Message from the content of t/samples/unsigned.eml
+my %sample;
+@sample{qw( file msg )} =
+    map +(
+        $_,
+        Sympa::Message->new_from_file($_)
+    ), "t/samples/unsigned.eml";
+
+# store a message and returns a hashref with its
+# qw( total_packets marshalled ) as keys.
+
+sub ok_bulk_store {
+    my ( $bulk, $msg ) = @_;
+    my $stored =
+        $bulk->store
+        ($msg
+        , 'dave.null@example.com' );
+    isa_ok $stored, HASH =>  '$bulk->store';
+    # TODO: skip them if $stored isn't HASH
+    ok +( 2 ==
+            grep
+            +(exists $$stored{$_})
+            , qw( total_packets marshalled ) )
+        , "got total_packets and marshalled from stored msg";
+
+    ok +( 2 == keys %$stored )
+        => "no extra information";
+    # TODO: test data extractions from marshalled name ?
+    $stored;
+}
+
+sub ok_no_next_from_empty_bulk {
+    my ( $bulk, $desc ) = @_;
+    my ( $msg, $file ) = $bulk->next;
+    ok +( not defined $file ) => "no next message $desc"
+        or note "next: $file";
+}
+
+sub ok_next_message {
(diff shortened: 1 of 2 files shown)


t/bulk.t
```diff
@@ -0,0 +1,167 @@
...
+{ no warnings;
+    *Sympa::Log::syslog = 1 #  $ENV{SYMPA_TEST_SILENT_SYSLOG}
+    ? sub {}
+    : sub {
+        shift; # $self isn't expected anymore
+        my ( $level, $fmt, @params ) = @_;
+        note sprintf "syslog $level: $fmt", @params; }
+    ;
+}
+
```
ikedas: I feel reflection would be better to be used as little as possible, and initializing Sympa::Log instance with usual way (setting {log_to_stderr} will output logs to STDERR).

t/bulk.t
```diff
@@ -0,0 +1,167 @@
...
+    # Sympa::Bulk::store needs
+    sympa_packet_priority => 5,
+    sympa_priority        => 1,
+    email                 => 'sympa',
+    domain                => 'lists.example.com',
+    sender_headers        => 'Resent-From,From,Return-Path',
+    list_check_suffixes   => 'list_check_suffixes',
+    umask                 => '027',
+);
+
```
ikedas: I can't understand why some of parameters above are assigned non-default values (especially, value for list_check_suffixes is obviously illegal).

t/bulk.t
```diff
@@ -0,0 +1,167 @@
...
+# as we are already in $self->{pct_directory}, -f will always fail with a
+# relative path.
+#
+# i (eiro) don't know if it's a bug or a feature
+
+use Cwd;
+$Conf::Conf{queuebulk} =
+    Cwd::abs_path
+    Sympa::Constants::SPOOLDIR."/bulk";
+
```
ikedas: Directorie paths defined in Sympa::Constants may be assumed to be absolute path.  (Except in the test cases explicitly assuming definition of relative paths,) SPOOLDIR should have been defined as absolute path in Sympa::Constants, and fix up here seems inappropriate.

t/bulk.t
```diff
@@ -0,0 +1,167 @@
...
+}
+
+# create a new Sympa::Message from the content of t/samples/unsigned.eml
+my %sample;
+@sample{qw( file msg )} =
+    map +(
+        $_,
+        Sympa::Message->new_from_file($_)
+    ), "t/samples/unsigned.eml";
+
```
ikedas: Path of test data is hardcoded, i.e. procedure to load test data is not generalized.

ikedas:
I added comments as I noticed.

Anyways, this is a test suite: This may be useful for _automating test_, however this itself will not _perform meaningful tests_.  Because test cases have not been written.

So I think this PR may not be rejected by negative reasons, but this PR does not have to be merged at present because of locking test.  I recommend submitter to close this PR when it will have been received comments, and to take enough amount of time to think out test cases for sufficient coverrage of test.

eiro:
hello @ikedas and thanks for reviewing ... few is better than nothing so i feel you should merge this test suite because it's a base to be improved and commented later. i'll comment your comment

t/bulk.t
```diff
@@ -0,0 +1,167 @@
...
+#                    !/,lock/
+#                and !m{(?:\A|/)(?:\.|T\.|BAD-)}
+#                and -f ($self->{pct_directory} . '/' . $_)
+#        } glob $self->{_glob_pattern}
+#    ];
+#
+# as we are already in $self->{pct_directory}, -f will always fail with a
+# relative path.
+#
+# i (eiro) don't know if it's a bug or a feature
```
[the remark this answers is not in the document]
eiro: i don't understand what you expected by this comment: do you want me to explain why your code fail with a relative path or do you want me to rewrite the comment to explain better why it fail with relative path? (i was expecting the testers to read the code there but the comment probably should avoid it and i can be more explicit)

t/bulk.t
eiro: i did that to make the syslog strings appears to the test report and i don't know if it's possible from the test framework itself. any idea to do that ?

t/bulk.t
eiro: list_check_suffixes is clearly a typo! thanks for spotting it. the others are stolen from the ConfDefs.pm AFAIR. do you want me to check all of them to be the default ones (assuming defaults comes from ConfDefs.pm) ?

t/bulk.t
eiro: i wrote it this way to keep a trace on the (i think undocumented) fact that those directories *must* be absolute. i'll check the doc on it

ldidry:
I :+1: @eiro on merging it even if some tests are lacking: having incomplete test suite for a subject is more encouraging for people to add tests than having no tests at all.

As a contributor, I can say "Oh, I can at least add a test, it won't be a big thing, but I can do it quick and it will help a little" instead of "Oh, man, I need to do all the writing. It's boring and will take long, I don't have time for that".

I suggests adding a warning saying "Please note that bulk test suite is not complete." when running the tests.

ikedas:
@ldidry, I don’t oppose to anyone merging this PR, but I won’t do.  Because I think a test suite not including nontrivial test cases is useless.

Why I think so is perhaps that, I have seen several *.t files including cases simply calling APIs by each, i.e. cases already known to succeed, have been kept without update: A few of them have been removed after all.

eiro:
I have no objection to see this PR closed for good reasons and being useless is a very good reason :).  so @ikedas can you backup what you said by pointing tests that already does a better job or describing what should be a better test suite ? (or writing it by yourself if you want to ?)

eiro:
code modified to take @ikedas comments in count

t/bulk.t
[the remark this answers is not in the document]
ikedas: I didn't think -f will fail with a relative path (` -f $_`), and I couldn't figure out what is the suspicious for you and why you noted it in test suite.

t/bulk.t
ikedas: Either using `note()` or STDERR, output seem to be hidden by test harness.  So I didn't figure out why you needed large-scale gimmick to use former way (`note()`).

t/bulk.t
ikedas: Test would be performed under controlled conditions. Otherwise, it might be hard to distinct unexpected result by a bug and unexpected result by randomly chosen condition.  Non-specific value is prefered for parameters unless there were reason to adopt special value.

dverdin:
This PR covers the basics of Bulk.pm, which was never done before. It is clearly an improvement.
This work should be later continued to introduce error test cases, such as badly formatted message and other potential errors but is is a very good start.
If I get a vote, I agree to merge this PR.

racke:
Grant me the time to review this PR properly. For certain typos should be fixed and the first note block should be removed - the latter belongs rather to PR comments than in the code ...

racke:
So test fails when running it as the other tests in travis.yml (where this one definitely belongs to), so I think it is premature to think about merging it.

```
racke@belukha:~/sympa-community/sympa$ make check-local TEST_FILES="t/bulk.t" 
[ -z "t/bulk.t" ] && TEST_FILES="t/01_Conf.t t/02_Sympa_DatabaseManager.t t/Language.t t/LockedFile.t t/Regexps.t t/compile_modules.t t/compile_executables.t t/Database_LDAP.t t/parse_templates.t t/pod-syntax.t t/tools_data.t t/tools_file.t t/Tools_Password.t t/Tools_SMIME.t t/Tools_Text.t t/tools_time.t"; \
PERL5LIB=src/lib; export PERL5LIB; \
/home/racke/perl5/perlbrew/perls/perl-5.24.0/bin/perl -MTest::Harness -e 'runtests @ARGV' $TEST_FILES
t/bulk.t .. 1/? Use of uninitialized value $dir in opendir at src/lib/Sympa/Tools/File.pm line 88, <DATA> line 1.
Use of uninitialized value $_ in -e at t/bulk.t line 102, <DATA> line 1.
Use of uninitialized value $Conf::Conf{"queuebulk"} in concatenation (.) or string at src/lib/Sympa/Bulk.pm line 56.
Use of uninitialized value $Conf::Conf{"queuebulk"} in concatenation (.) or string at src/lib/Sympa/Bulk.pm line 56.
Use of uninitialized value $Conf::Conf{"queuebulk"} in concatenation (.) or string at src/lib/Sympa/Bulk.pm line 56.
Use of uninitialized value $Conf::Conf{"queuebulk"} in concatenation (.) or string at src/lib/Sympa/Bulk.pm line 56.
Use of uninitialized value $Conf::Conf{"queuebulk"} in concatenation (.) or string at src/lib/Sympa/Bulk.pm line 56.
Use of uninitialized value $directory in -d at src/lib/Sympa/Bulk.pm line 82.
Use of uninitialized value $directory in mkdir at src/lib/Sympa/Bulk.pm line 84.
Use of uninitialized value $directory in sprintf at src/lib/Sympa/Bulk.pm line 92.
Cannot create : No such file or directory at src/lib/Sympa/Bulk.pm line 92.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 2 just after 2.
t/bulk.t .. Dubious, test returned 2 (wstat 512, 0x200)
All 2 subtests passed 

```

ikedas:
I confess that I have reviewed this PR only on desktop (and I haven’t clearly wrote on this parameter in POD of Bulk.pm: those are my failure).

Anyway, if I was asked, I’ll deny merging this PR written for a few tens of minutes in weekend evening.

I want to ask submitter again for spending enough time to read sources, run software as necessity, understand what they do, and think out enough coverage of test cases. Test suite may come at the last.

t/bulk.t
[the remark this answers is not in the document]
racke: Testers aren't supposed to read the files ... this is mostly our job. If the relative path is actually a problem, it should be reported as issue in itself IMHO.

t/bulk.t
racke: There are test modules to capture logging. I think that defeats the principle of minimum of changes and is hard to read as well.

t/bulk.t
racke: At any rate, the "use Cwd" belongs to the top of the file, there is no point in:

- break it up in multiple lines
- skip the parenthesis
- use concat operator without whitespace

racke:
I also suggest to decline that pull request. Maybe @eiro could try to start fry small fish, e.g. hunt down missing POD entries like the one @ikedas referred to.

eiro:
hello @racke,

> There are test modules to capture logging. I think that defeats the
> principle of minimum of changes and is hard to read as well.

good idea to use an external module but i don't want to avoid
dependencies just writing a test. should be improved later.

marc

eiro:
hello,

> You can always skip a test (file) if dependencies are missing. Better
> to use known-to-work solid CPAN modules rather than homegrown code
> that is hard to understand and looks really hacky.

ok then. i'll find a module and fix it

regards

eiro:
> +use Cwd;
> +$Conf::Conf{queuebulk} =
> +    Cwd::abs_path
> +    Sympa::Constants::SPOOLDIR."/bulk";


> At any rate, the "use Cwd" belongs to the top of the file

it seems those lines should disapear or at least be moved at some point.
so i used Cwd there to be sure it will be removed too.

> there is no point in:
> - break it up in multiple lines

when everything is in a long line, it's hard to get what is what
immediatly

    $Conf::Conf{queuebulk} = Cwd::abs_path Sympa::Constants::SPOOLDIR . "/bulk";

> - skip the parenthesis

there is no point in adding them

> - use concat operator without whitespace

right! i'll fix it.

racke:
@eiro Really, this is *not* about your preferences - but what is commonly used. If you use a function outside of a DSL or really standard ones like print, please use parenthesis.

Also a function call with one parameter usually is fine on one line.

eiro:
hello,

> There is a lot of things missing to make it a good contribution for Sympa.

yes! thanks for helping me to improve it.

> +@sample{qw( file msg )} =
> +    map +(
> +        $_,
> +        Sympa::Message->new_from_file($_)
> +    ), "t/samples/unsigned.eml";
> +
> 
> Also that is terrible to read. No map needed for a list of one item.

ok then ... does

    my %sample = (
        file => "t/samples/unsigned.eml",
        msg  => Sympa::Message->new_from_file("t/samples/unsigned.eml")
    );

looks better for you?

> I really don't like that. Why do you assign something to $_ ? It looks like a simple test but it is hard to comprehend. 

right ... This was too cryptic and will be replaced.

> You don't use that variable below, right?

right!

again, thanks for reviewing

marc

eiro:
> Testers aren't supposed to read the files ... this is mostly our job.
> If the relative path is actually a problem, it should be reported as
> issue in itself IMHO.

any silent fail is a problem to me so i'll open an issue later.

thanks
marc

eiro:
hello,

based on your experience, can you advice any module/practice to capture
stderr as a diag/note?

marc

ikedas:
@eiro,

> based on your experience, can you advice any module/practice to capture
stderr as a diag/note?

My advice:

Oradinarily, here is the place for questions/answers.  However, here is _not_ the place to give answers to persons who won't think by themselves.

  - Before you ask the others for answer, have you looked for ansewer?  Always asking without investigating, you won't gain experiences.

    Don't you think that you ask the others before searching by yourself because you will get the answer quickly and easily, do you?

  - You would be better to say,
    > I have looked for modules/practices, and I found these things: (things you found). Among them, I think (A) is the best, because (reason you chose A).  What do you think?

    Couldn't you type "_module/practice to capture stderr as a diag/note_" into form of  search engine?  Couldn't you understand what shown in search results?

ikedas:
This pr no longer is in sync with recent codebase. Close for now.

PR #440, opened 2018-10-12
 Use tidyall to ease tidying files

ldidry:
Follow-up of #438 

- Code::TidyAll added as develop requirement in cpanfile
- Add tidyall target to Makefile.am (to run it with the good options
  without pain)
- Add dot/dot.tidyallrc
- Add t/ and xt/ to perltidy test

\+ Tidy up src/ t/ xt/ files

PR #445, opened 2018-10-17
Test: Make xt/perltidy.t to check *.pl.in and *.fcgi.in as well.

ikedas:
This may be an enhancement to PR #322.
See also #440 for detail.


racke:
Yeah, good idea. :+1:

PR #452, opened 2018-10-20
Implementing expire_members task

ikedas:
- [feature] Implementing "expire_members" task.
    Note: "expire" task has never been implemented and expire_task parameter was kept but ineffective. To avoid unintentional execution of task with old setting, new parameter for "expire_members" should not be an alias of old one.
  - [bug] next() function in task resets original creation_date. Fixed by renaming task file instead of removing it (That's why "expire" task did not work as expected).

This PR depends on PR #394.

### DIFF
--- a/default/Makefile.am
+++ b/default/Makefile.am
@@ -171,6 +171,12 @@ nobase_default_DATA = \
 	scenari/visibility.secret \
 	tasks/eval_bouncers.daily.task \
 	tasks/expire_bounce.daily.task \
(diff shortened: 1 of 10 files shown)


acasadoual:
Has expire_task never been implemented in 6.2.37b1?? It's allow set in Admin page of sympa.

racke:
I think that is the original problem which is supposed to be solved with this PR @acasadoual

ikedas:
I submitted new issue #475.

Issue #451, opened 2018-10-19
Optimization of web GUI / adding an index on prev_id_session DB field


salaun-urennes1:
While preparing our ugrade to Sympa 6.2.36 we noticed bad performances on every web request. I figured out that adding an index on prev_id_session DB field saves 1.2sec on every web request.

Version
-------
6.2.36

Installation method
-------------------
source package

While running NYTProf profiler I detected too much time spent on WWW::Session : 
![image](https://user-images.githubusercontent.com/4091149/47222822-ff452e80-d3b7-11e8-931c-cb64c4f82dc4.png)

I then found out the DBI requests where the culprits : 
![image](https://user-images.githubusercontent.com/4091149/47222899-33b8ea80-d3b8-11e8-8fbf-bf18963afa6f.png)

On our server the session_table has 464712 entries ; the SQL query to load a session takes 1.4sec.

I then added on index on prev_id_session DB field ; the same SQL query now takes 0.2sec

**=> I suggest you add an index on prev_id_session DB field**

racke:
It sounds reasonable to me to add an index to a field used very often in a WHERE condition.

salaun-urennes1:
I went further in my tests and it turns out that, even with this index added, web access are still between 2.5sec and 4.2sec (just to access a list homepage). 

I tried to empty the session_table and waooooh : access to the same web page takes from 0.4sec to 0.7sec :-)

=> I will do some further investigations to find out why the `purge_session_table_task` (default value for `session_table_ttl `is `2d`) is not doing its job properly...

racke:
Yes please do check that task - this should be really working reliable :roll_eyes:

salaun-urennes1:
That task did not run since 2017 dec 15th :-(
```
$ ls /usr/local/sympa/spool/task/*session*
/usr/local/sympa/spool/task/1513302372.ACTION.purge_session_table._global
```

I go on my investigations...

salaun-urennes1:
I found out why the `purge_session_table_task` task was not run on our 6.2.16 instance : task_manager process would die with the following error: `Undefined subroutine &Sympa::send_notfy_to_listmaster called at /usr/local/sympa-6.2.16/bin/task_manager.pl line 1925`

That bug has been fixed meanwhile.

But it would still make sense to add an index on prev_id_session DB field...

racke:
I agree with you on having an index on prev_id_session. Do you want to produce a PR for this change?

salaun-urennes1:
[...]

PR #460, opened 2018-10-31
Add index on prev_id_session DB field, see #451

salaun-urennes1:
Suggested on  #451

### DIFF
--- a/src/bin/create_db.Sybase
+++ b/src/bin/create_db.Sybase
@@ -212,6 +212,9 @@ create table session_table
 )
 go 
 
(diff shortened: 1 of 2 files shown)


racke:
We formally dropped support for Sybase (#417), so there is no need to patch create_db.Sybase. In fact, we might remove the remaining Sybase files from the sources.

Other than that, I like the patch :+1: and this is really good find of you. Thanks!

ikedas:
Merged as c77b07aa080cac5f3f30ef3fae26c3ae02680904

PR #471, opened 2018-11-09
Fix #323 — Add "Report abuse" button to lists side menu

### LINKED ISSUE #323: Feature: add a "report abuse" button on lists info page
- [x] create a template that will be included into lists info page
- [x] the template will simply be a mailto link to listmaster with subject "Sympa abuse report: list_name" (something like that)

ldidry:
- [X] create a template that is included into lists side menu (instead of just info page as said in #323, to be always visible, whatever the page of the list you visit)
- [X] the template is just a mailto link to listmaster with subject "Sympa abuse report: list_name"

### DIFF
--- a/default/web_tt2/report_abuse.tt2
+++ b/default/web_tt2/report_abuse.tt2
@@ -1,0 +1,3 @@
+<!-- report_abuse.tt2 -->
+[% | mailto("${conf.listmaster_email}@${domain}",{subject=>"Sympa abuse report: ${list}"}) %][%|loc%]Report abuse[%END%][%END%]
+<!-- end report_abuse.tt2 -->
--- a/default/Makefile.am
+++ b/default/Makefile.am
@@ -321,6 +321,7 @@ nobase_default_DATA = \
 	web_tt2/remove_arc.tt2 \
 	web_tt2/rename_list_request.tt2 \
 	web_tt2/renewpasswd.tt2 \
+	web_tt2/report_abuse.tt2 \
 	web_tt2/requestpasswd.tt2 \
 	web_tt2/request_topic.tt2 \
 	web_tt2/reviewbouncing.tt2 \
--- a/default/web_tt2/list_menu.tt2
+++ b/default/web_tt2/list_menu.tt2
@@ -156,4 +156,10 @@
 [% PROCESS additional_list_menu_links.tt2 %]
 [% CATCH %]
 [% END %]
+
+[% IF conf.show_report_abuse %]
+    <li>
+        [% PROCESS report_abuse.tt2 ~%]
+    </li>
+[% END %]
 <!-- end list_menu.tt2 -->
--- a/src/cgi/wwsympa.fcgi.in
+++ b/src/cgi/wwsympa.fcgi.in
@@ -1146,6 +1146,7 @@ while ($query = CGI::Fast->new) {
         'automatic_list_families',
         'spam_protection',
         'pictures_max_size',
+        'show_report_abuse',
     ) {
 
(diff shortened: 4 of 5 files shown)


ldidry:
It's a generic report link. The [use case described](https://github.com/sympa-community/sympa/issues/323#issuecomment-392243687) in the issue was list-centered.

I think it's the reporter that should provide details in the mail body.

xavierba:
Fair enough. Just take my comment as a nice to have addition to the feature. Having something to report abuse, even if not perfect, is better than nothing, indeed :-)

default/web_tt2/report_abuse.tt2
```diff
@@ -0,0 +1,5 @@
+<!-- report_abuse.tt2 -->
+[% IF conf.show_report_abuse %]
+    [% | mailto("${conf.listmaster_email}@${domain}",{subject=>"Sympa abuse report: ${list}"}) %][%|loc%]Report abuse[%END%][%END%]
+[% END %]
+<!-- end report_abuse.tt2 -->
```
ikedas: `[% IF ... END %]` is redundant: One in `list_menu.tt2` is enough.

po/sympa/sympa.pot
```diff
@@ -6876,6 +6884,10 @@ msgstr ""
 msgid "Request new password"
 msgstr ""
 
+#: default/web_tt2/report_abuse.tt2:3
+msgid "Report abuse"
+msgstr ""
+
```
ikedas: `*.pot` file need not updating: It will be done during release process.

default/web_tt2/report_abuse.tt2
ikedas: The subject `Sympa abuse report: ${list}` would be better to be translated.

ldidry: Mmh… not sure: on Framalistes (french instance), we have german-speaking people using our instance. I would receive a mail with a subject in german, and so would not give it with all the attention it deserves + there's a risk that I may think it's a spam. 

I'm open to discussion, though. Maybe translating it to the instance default language? Can I force the translation to a given language? If yes, how?

po/sympa/sympa.pot
ldidry: I didn't knew that, sorry.

ikedas: Don't worry. I'm glad now you know about common job.

default/web_tt2/report_abuse.tt2
ikedas: I feel it making sense that we stick to single subject (language) so that we won't miss the report.  If it would be translated, language of domain that listmaster belongs to would be used.

However, currently it is difficult for templates to choose language differ from list's lang.  Maybe sticking to english is reasonable.

Additonally, I found that [syntax of mailto URL](https://tools.ietf.org/html/rfc6068#section-2) containing non-ASCII parameters can be complex and it might mess several mailers.  Limiting characters in subject to ASCII looks reasonable.

ldidry: I first wanted to put the if in `report_abuse.tt2` then noticed than I needed one in `list_menu.tt2` (because of the `<li>`) but forgot to remove it here.

Fixed now.

po/sympa/sympa.pot
ldidry: Removed from the commit.

default/web_tt2/report_abuse.tt2
ldidry: I agree.

ikedas:
@ldidry , thanks for resolving! If no objection in this week, I'll merge it (or anyone else may merge).

Issue #296, opened 2018-05-15
Feature request: domains correction


ldidry:
Some people subscribe to list with well-know mistakes (ie foo@gmail.fr instead of foo@gmail.com).

I suggest to have a setting to automatically correct those mistakes, like an array or an hash of domains and there correction, maybe with globbing (not regex, too error-prone) to be able to set gmail.* => gmail.com.

ldidry:
The current PR need some advices from the community. See the discusson on sympa-developpers  ML: https://listes.renater.fr/sympa/arc/sympa-developpers/2018-11/msg00008.html

ping @racke @dverdin?

PR #478, opened 2018-11-16
Fix #295 — Add domain blacklist feature

### LINKED ISSUE #295: Feature request: domains blacklist
Some people migrating lists from <whatever, I don't know> to our Sympa registers people with some well-known abandoned addresses. In France, we got `voila.fr` that is still registered but without mail server (and it will not change, it's owned by a big ISP that shut down the mail server a few years ago).

I suggest a domains blacklist setting that will prevents addresses using those domains to subscribe to list (a message saying "Sorry, that domain is blacklisted" to the subscribing action instigator could be cool)

ldidry:


ldidry:
I'm not 100% sure that there is no other files that can add subscribers. Need a good review.

ikedas:
This may fix #295.

ikedas:
You are adding a test against domain blacklist to `add` handler.  Probably `subscribe` handler may also have such test.

ldidry:
Thanks, I added the test in `subscribe`. Since I use more or less the same test in 3 files, I create `Sympa::Tools::Domains` and put a function to do the test in it. I will extend it for #296

dverdin:
I arrive a bit late but did you know that Sympa can do this already?
for each list, you can set a blacklist. If your blacklist entry looks like "*@domain.tld", you blacklist the whole domain.
This blacklist is automatically used by any Sympa list.
It can be defined at the server or virtual host level, as it is a scenario.
You could also have used implicit scenarios inclusion.
In other words: you already had solutions in Sympa without the hasle of writing new code.

Remember: most of the time, any problem in Sympa can be solved by a scenario.

ikedas:
@dverdin, you are right. However I think
  1. current feature cannot have lists of blacklisted users by each action (scenario function), and
  2. use of wildcard character is not always intuitive for all users.

dverdin:
Hi @ikedas ! And a happy new year ! And I added the 6.2.38 to the sympa.org web site!

Now, back to the topic of this PR:
1. Actually, the blacklist can also be applied to subscription.
2. Maybe but it is explained just above the blacklist. Honestly, it does not look that hard.
And Finally, that looks more like a listmaster problem. The issue quoted wants to prevent specific domains from being added when people import lists. The people importing such addresses will certainly not take the time to edit the blacklist. Hence my proposal to use an implicitely included scenario.

ikedas:
@dverdin, if you want to raise more discussion, could you please open a new issue?  This PR has already been approved, merged and closed as the goal of an feature request submitted more than half a year ago.

dverdin:
@ikedas : You're right. I'll open a feature about it.
See you!

dverdin:
I completed the original feature request.

Issue #72, opened 2017-10-02
sympa init script on Debian [: =: unary operator expected


pkissman:
When you install a new version of sympa 6.20 or 6.22,  the initscript generated has the following warning on a Debian system (wheezy):
  **/etc/init.d/sympa: line 262: [: =: unary operator expected**

Adding quotes around the NETWORK reference in if [ ${NETWORKING} = "no" ]  in sympa-6.2.22/src/etc/script/sympa.in

if [ **"**${NETWORKING}**"** = "no" ]

seems to do the trick

racke:
Thanks for the report @pkissman. But I think we should stop supporting init scripts as modern distributions are using systemd. Not that I really in love with that ... 

Given the small size of the current Sympa dev team, I don't see a point to support Debian wheezy etc.

pkissman:
That's fine.  I can fix it myself at each upgrade install, which I do from source, not from  Debian packages.  We hope to hop up to a current OS distribution soon any.

Thanks for considering.

pjk



________________________________

PR #479, opened 2018-11-17
Fix missing quotes in init script

rseichter:
Issue https://github.com/sympa-community/sympa/issues/72 states that "we should stop supporting init scripts as modern distributions are using systemd". I have to disagree. Gentoo Linux defaults to OpenRC, so I created a small pull request to fix missing quotes in the init script.


racke:
Well I said that ... but on the pretext that we can't handle it on our own. If others support it, we are glad about that. Thanks for the patch!

PR #483, opened 2018-11-28
Fix typo in web_tt2/confirm_action.tt2

cgx:
## Description

There's a typo in a localized string of `confirm_action.tt2`.

## Impacts

This PR fixes the typo that appears in the template and all the localization files.


ikedas:
[Fixed](https://translate.sympa.org/en_US/sympa/translate/en_US.po#unit=59576).

Issue #461, opened 2018-10-31
Template parsing problems in parameterizable data sources


junior013:
Version
-------
6.2.24

Installation method
-------------------
deb (Ubuntu 18.04 LTS, bionic repo)

Expected behavior
-----------------
Import members from predefined LDAP with filter as a parameter

Actual behavior
---------------
- If source_parameters defined -> Works
- If source_parameters left empty -> The unparsed template code goes in the LDAP query

Additional information
----------------------
Template V1:
`filter (&(cn=*)[% param.0 %])`
log:
`wwsympa[6381]: err main::#1606 > main::do_review#4842 > Sympa::List::on_the_fly_sync_include#6929 > Sympa::List::sync_include#6581 > Sympa::List::_load_list_members_from_include#5783 > Sympa::List::_include_users_ldap#5059 > Sympa::DatabaseDriver::LDAP::do_operation#204 Unable to perform LDAP operation: Bad filter`
`wwsympa[6381]: err main::#1606 > main::do_review#4842 > Sympa::List::on_the_fly_sync_include#6929 > Sympa::List::sync_include#6581 > Sympa::List::_load_list_members_from_include#5783 > Sympa::List::_include_users_ldap#5071 LDAP search (single level) failed: (89) Bad filter (searching on server X:389; suffix X; filter (&(cn=*)[% param.0 %]); attrs: mail)`

Template V2:
`[% SET filter = '' %]`
`[% IF param.defined AND param.defined(0); filter = param.0; ELSE; filter = 'cn=*'; END %]`
`filter [% filter %]`
log:
`wwsympa[6381]: info Sympa::List::_load_include_admin_user_file() Bad line "[% SET filter = '' %]#012" in /etc/sympa/data_sources/ad_users.incl`
`wwsympa[6381]: info Sympa::List::_load_include_admin_user_file() Bad entry "[% SET filter = '' %]" for key "ALL", paragraph "include_ldap_query" in /etc/sympa/data_sources/ad_users.incl`
`wwsympa[6381]: info Sympa::List::_load_include_admin_user_file() Bad line "[% IF param.defined AND param.defined(0); filter = param.0; ELSE; filter = 'cn=*'; END %]#012" in /etc/sympa/data_sources/ad_users.incl`
[...]

PR #486, opened 2018-11-30
Fix code tidying

ldidry:


ikedas:
@ldidry, I run “make tidyall” just before 6.2.37b.2.  Do you know why things went wrong? (Though I have no objection to merge this fix soon.)

ldidry:
I was about to say "Well, there's 29 commits between 6.2.37b.2 and current sympa-6.2 branch, I think that explains that.", but with a [git blame](https://github.com/sympa-community/sympa/blame/sympa-6.2/src/lib/Sympa/List.pm#L5796) on `src/lib/Sympa/List.pm`, it seems that a untidyed block of code has been there for 4 years.

I guess that's a glitch in perltidy.

I noticed today that perltidy can say a thing on a test, and find new unrelated things to tidy up after correction from the previous test. Some code may have been added that triggers perltidy like that.

ikedas:
Perltidy won't take the blame: I found it was due to quick fix for #461.
Anyway it should be merged.

PR #488, opened 2018-12-02
Prevent warning on undefined salt variable.

racke:


ikedas:
This is introduced on PR #398. I guess this may be related to issue #167 (not surely).
I wish it would be kept before release of 6.2.38.

PR #492, opened 2018-12-05
Add token to prevent CSRF.

racke:
This is a slightly adjusted version of the CSRF patch provided by @mpkut. I tested in my working copy over the weekend and I didn't run into a regression. This is scheduled to be part of the next beta release, as agreed with @ikedas.

### DIFF
--- a/default/web_tt2/admin.tt2
+++ b/default/web_tt2/admin.tt2
@@ -29,6 +29,7 @@
 [% IF is_privileged_owner %]        
   [% IF list_conf.status == 'closed' ~%]
     <form name="manage_list_status" action="[% path_cgi %]" method="post">
+    <input type="hidden" name="csrftoken" value="[% csrftoken %]" />
     <div>
     <input class="MainMenuLinks" type="submit" name="action_open_list"
       value="[%|loc%]Restore List[%END%]" />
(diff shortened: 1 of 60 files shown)


ikedas:
@mpkut & @racke, thanks much for enhancement & testing!

- `renewpasswd.tt2` does not have `csrftoken` field.  Is this intended? (I guess this does not matter anyway.)
- I realized that users may have to update many web templates they have customized, and I prepared [additional changes](https://github.com/racke/sympa/compare/pr/csrftoken...ikedas:racke/pr/csrftoken) to avoid updating templates.  Is this useful?

mpkut:
Hello @ikedas,

* You're right, `renewpasswd.tt2` does not really need a `csrftoken` field because it does not employ the authorized user's session to make any state changes.
* Your additional changes are a *much* more airtight and maintainable solution to the question of how to consistently supply the token. I had been concerned about adding another field to all of those forms, but simply did not have the imagination to realize that your approach was a possibility.

Thank to you and Racke for your review and helping arrive at a well-formed solution to this concern!

racke:
Your approach sounds good to me, @ikedas

ikedas:
ok. I submitted #493. It may be merged in beta.3 or 6.2.38.

PR #494, opened 2018-12-07
Fix #301 — Add global mail signature feature

### LINKED ISSUE #301: Feature request: add a "global" mail signature
On 6.2.16, you can't globally add a mail signature set by the admins with lists that already have a list signature, you can just set the default mail signature.

As we wanted to always add an unsubscribe link to every mail of every list (to avoid the "Please unsubscribe me" mail that people would have sent to us), we removed the possibility for list owner to change the mail signature.

Sympa should allow to set a "global" mail signature that would be added to every mail of every list in addition to the lists signature.

ldidry:
Add a etc/mail_tt2/message.global_footer or etc/mail_tt2/message.global_footer.mime file and it will be added to all messages.
Obviously, only admins can edit that file.
It can be used to add an unsubscribe link while letting users add their own list signature.

Will fix #301

### DIFF
--- a/src/lib/Sympa/Message.pm
+++ b/src/lib/Sympa/Message.pm
@@ -1693,14 +1693,31 @@ sub _decorate_parts {
         }
     }
 
+    my $global_footer;
+    foreach my $file (
+        $Conf::Conf{'etc'} . '/mail_tt2/message.global_footer',
+        $Conf::Conf{'etc'} . '/mail_tt2/message.global_footer.mime'
+    ) {
+        if (-f $file) {
+            unless (-r $file) {
+                $log->syslog('notice', 'Cannot read %s', $file);
+                next;
+            }
+            $global_footer = $file;
+            last;
+        }
+    }
+
     ## No footer/header
-    unless (($footer and -s $footer) or ($header and -s $header)) {
+    unless (($header and -s $header)
+        or ($footer and -s $footer)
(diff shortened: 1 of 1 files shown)


ValentinDrean:
Hi,

My instance version is 6.2.40 and I still can't make this global footer feature work on my lists where there is no footers set. 

The idea is to create a default footer file when no footer have been set on a list in this path here : `/var/lib/sympa/expl/list-instance/list-name/message.footer`, I have more than a thousand lists where not all have an unsubscribre link in footer so it's not really « RGPD friendly » ;\

Web feature (`https://list-site.tld/wws/serveradmin/templates`) to change default/global footer does not work either, when I save it goes back blank/empty again.

I tried directly editing several configuration files to add these `message.global_footer` and `message.global_footer.mime` in :

- `/usr/share/sympa/default/mail_tt2/` (extension files here are `.tt2`)
- `/etc/sympa/mail_tt2`
- `/etc/sympa/list-site.tld/mail_tt2`

`merge_feature` is `on` in my  `/etc/sympa/sympa/sympa.conf`

I'm not very experienced as I don't know which logs and other conf files I need to edit to make this global footer setting work /o\

This seemed a right place to ask but sorry if this is not, maybe I'll double this tonight on renater support list I guess, thanks for all that crazy work on Sympa anyway <3

racke:
@ValentinDrean I understand your problem, but please open a fresh issue.

PR #504, opened 2018-12-12
WWSympa: Loading home page takes long time

ikedas:
When user is logged-in, Sympa collects information about membership/adminship among all lists, content of moderation spool and so on, though they will _never_ be used.

Fixed by removing code collectiong information in vain. Deprecated web_tt2/news.tt2.


racke:
Looks like a good idea :+1:

PR #511, opened 2018-12-21
Fix #300 — Allow to delete your own account

### LINKED ISSUE #300: Feature request: a "delete my account" button
There should be a "delete my account" button that unsubscribe you from all lists and remove your account immediatly.

To protect that feature, Sympa could
1. asks for the user password
1. send a mail with a link containing a token

ldidry:
Only from web interface for now.

- add a form on pref page to delete the account
- checks the password of the user
- if it's good:
  - unsubscribe the user from all its lists
  - remove the user from the editor role of all its lists
  - remove the user from owner role of all its lists *only* if he/she
    is not the only owner (=> don't delete the account and give an error message)
  - if he/she was the only privileged owner of a list, give privileged
    profile to all unprivileged owners

### DIFF
--- a/default/mail_tt2/report.tt2
+++ b/default/mail_tt2/report.tt2
@@ -836,6 +836,9 @@ Warning: this message may already have been sent by one of the list's moderators
 [%~ ELSIF report_entry == 'still_owner' ~%]
   [%|loc(report_param.lists)%]You are the only owner of the following list(s): %1. Please give ownership to other people before deleting your account. You have been unsubscribed from all your lists though.[%END%]
 
+[%~ ELSIF report_entry == 'no_classic_session' ~%]
+  [%|loc()%]You are not authorized to delete your account if you are not using the built-in authentication (i.e. you are using a LDAP authentication, a SSO system, etc.).[%END%]
+
 [%~ END ~%]
 
 [%~ END ~%]
--- a/default/mail_tt2/report.tt2
+++ b/default/mail_tt2/report.tt2
@@ -438,6 +438,9 @@
 [%~ ELSIF report_entry == 'logout' ~%]
   [%|loc%]You have logged out[%END%]
 
+[%~ ELSIF report_entry == 'account_deleted' ~%]
+  [%|loc%]You have been unsubscribed from all your lists and your account has been deleted[%END%]
+
 [%~ END ~%]
 
 [%################~%]
@@ -829,6 +832,9 @@ Warning: this message may already have been sent by one of the list's moderators
 
 [%~ ELSIF report_entry == 'owner_domain_min' ~%]
   [%|loc(report_param.value,report_param.owner_domain_min,report_param.owner_domain)%]Unable to reduce the number of list owners in required domains to %1. Domains that count toward the minimum requirement of %2: %3[%END%]
+
+[%~ ELSIF report_entry == 'still_owner' ~%]
+  [%|loc(report_param.lists)%]You are the only owner of the following list(s): %1. Please give ownership to other people before deleting your account. You have been unsubscribed from all your lists though.[%END%]
 
 [%~ END ~%]
 
--- a/default/web_tt2/confirm_action.tt2
+++ b/default/web_tt2/confirm_action.tt2
@@ -1,4 +1,7 @@
 <!-- confirm_action.tt2 -->
+[%#
+  # Set h2 title and text of confirmation page
+  #%]
 [% IF confirm_action == 'add' ~%]
   [% IF previous_action == 'show_exclude' ~%]
     <h2><i class="fa fa-check-circle"></i>
@@ -78,6 +81,15 @@
   </h2>
   <p><strong>
     [%|loc%]Do you really want to unsubscribe ALL selected subscribers?[%END%]
+  </strong></p>
+[%~ ELSIF confirm_action == 'delete_account' ~%]
+  <h2><i class="fa fa-check-circle"></i>
+    [%|loc%]Delete my account[%END%]
+  </h2>
+  <p><strong>
+    [%|loc%]Do you really want to unsubscribe you from all your lists, remove your ownership of your lists and permanently delete your account?[%END%]
+    <br />
+    [%|loc%]Please note that you will not be able to delete your account if you are the only owner of one or more list.[%END%]
   </strong></p>
 [%~ ELSIF confirm_action == 'auth_add' ~%]
   <h2><i class="fa fa-check-circle"></i>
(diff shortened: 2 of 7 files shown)


ikedas:
> Only from web interface for now.

  * I think this doesn't matter. Authentication with web interface is almost only way for Sympa to identify the users certainly.

> * add a form on pref page to delete the account
> * checks the password of the user
> * if it's good:
> 
>   * unsubscribe the user from all its lists
>   * remove the user from the editor role of all its lists
>   * remove the user from owner role of all its lists _only_ if he/she
>     is not the only owner (=> don't delete the account and give an error message)
>   * if he/she was the only privileged owner of a list, give privileged
>     profile to all unprivileged owners

  * Between passwrod verification and actual deletions, confirmation (e.g. using `$session->confirm_action`) seems better.

Additional comment:

  * This feature allows users to delete their personal information in an identity source `user_table`.  Thus, IMO it should be enabled when the user is identified by this source and disabled when they are identified by the other sources such as LDAP, SSO.  --- In short, this feature may be available when user has logged-in with built-in authentication (`$session->{auth} is `'classic'`).

ldidry:
- [X] Between passwrod verification and actual deletions, confirmation (e.g. using $session->confirm_action) seems better.

I ask the password on the pref page, then ask confirmation, is it what you meant?

ldidry:
- [X] disabled when they are identified by the other sources such as LDAP, SSO

dverdin:
In order to allow it from other sources (SOAP or REST for example, but it could also be resuested by mail and trigger an email confirmation), wouldn't it be nice to have the code in a Request Handler, the way we will do for list admins management? (See  #524)
Just saying, because having web-specific queries have often be a cause of troubles later. When we want to generalize a mechanism.
If it's too much work, I can try do to it later.

dverdin:
In addition, that should also respect the unsubscribe scenario. If unsubscription is prevented, they should not be able to do it.

dverdin:
I did not get why it is disabled when identified through a SSO ?

racke:
@dverdin  Sympa doesn't control the SSO, LDAP, ... ?

dverdin:
@racke : Yes, I get that but the SSO is just a way to authenticate to Sympa. Being authenticated does not mean you have any subscription or handle any list.
Therefore, using an SSO should not prevent you from clearing you subscriptions.
Only authorization scenarios should be used to allow / prevent actions. I think the "signoff" scenario is the best candidate for this: if you are allowed to unsusbcribe, you can do it. If you are not, you can't.
The handling of list ownership which blocks if no owner remains looks OK to me.
In addition, we could also introduce a simple robot parameter which would enable / disable this feature globally.

ikedas:
> @racke : Yes, I get that but the SSO is just a way to authenticate to Sympa. Being authenticated does not mean you have any subscription or handle any list.
> Therefore, using an SSO should not prevent you from clearing you subscriptions.
> Only authorization scenarios should be used to allow / prevent actions. I think the "signoff" scenario is the best candidate for this: if you are allowed to unsusbcribe, you can do it. If you are not, you can't.
> The handling of list ownership which blocks if no owner remains looks OK to me.
> In addition, we could also introduce a simple robot parameter which would enable / disable this feature globally.

Let's clarify my argument:

Subscription and adminship on the lists are matter of authorization. Managing account on a system is matter of authentication.

If this PR aims to implement measure to clear whole account information of a person on a Sympa server, authentication matters: A person authenticated through one identity source (e.g. LDAP) _must not_ control another identity source (e.g. user_table), and vice versa.  Otherwise, the weakest authnetication won.

racke:
I second @ikedas's argument.

dverdin:
If it is the case, I still don't see.
I know there is not the same level of assurance between authentication systems, but I still don't see why hard-code restrictions to some actions based on authentication mechanisms. It looks more like an administrator matter than a developper's.
There is certainly something I did not understand. I'll re-read the code and come back.

racke:
For subscribers we have a mechanism to override the list from external sources (exclusion_table), so if the subscriber unsubscribes from a list Sympa adds an entry to a table.

For users we don't have such a mechanism as far I know, so if the user is delete from the Sympa database the account will recreated next time the user does a successful login based on the external source.

dverdin:
The account will be restored. The entry in the user table. But not the lists memberships/ownerships.

dverdin:
If a user chooses to base authentication on another system, it's his business, not ours.
A lot of services use SSO as a convenience because they use several system in addition to Sympa. The user could still want to remove all hist memberships / ownerships.
And what about those which have several authentication mechanisms?

racke:
@dverdin Let's stop this fruitless discussion. The feature to remove accounts from external sources can be added later. Please stick to the principle of minimal change sets.

dverdin:
@racke : I'm exactly trying to keep the change set minimal by preventing the meaningless exception added about SSO.
Deleting from the user_table is different from removing all subscriptions and ownerships.
If we can't discuss it here, when the change is about to be done, then there is no point in discussing at all.

So I'm telling it again for the last time: if we restrict the ability to remove all subscriptions and list ownerships on the basis of the authentication method, we add a pointless exception, that we'll need to fix later.

racke:
"Deleting your account" should effectively do that, which isn't really true in case of external authentication. And the user is not in charge what kind of authentication he *has* to use. That's my point and vote for accepting the PR with the external authentication exception.

dverdin:
But when somebody wants to remove his account, he means the subscriptions and ownerships. He means the actual data he manipulates. The point is to stop receiving mails and notifications. Leave a community.
Preventing someone from doing it because there might remain an entry in the database it the user logs in again is clearly wrong.
You know the difference, in Sympa, between the user table and the subscriptions, don't you?
You also know that the user table is automatically cleaned from entries for whoch no subscriptions remain?
So you know that the worse that could happen would be that the user's email address would come back, for a limited time, in the database.  And we can address it later.
In the meantime, I vote against the exception of preventing this cleanup based on how the user authenticated.

ikedas:
This page aims for review on a PR.  If you want to discuss about the ideas motivating it, please continue on the page of original feature request #300.

dverdin:
These are my arguments for why the commit https://github.com/sympa-community/sympa/pull/511/commits/6ffdd64d03e68e9d3043d6f429737f9f235aec70 is wrong in this PR.

dverdin:
Erm. I see what's wrong.
With a confirmation using the password, the solution as it is written can't be applied to third party authentication.
OK. Let's compromise: You apply the PR as it is and I fix it after by:
- allowing third party authentication,
- replace password confirmation by email confirmation,
- add control to the feature with:
    - a global parameter to enable / disable it
    - add control with the unsubscribe scenario.

I'll propose this in the issue.

ikedas:
This PR fixes #300.

ilf:
Thanks for this issue and the implementation!

It's been two years, maybe we can finish the "experimental" phase.

I propose to enable this by default. Allowing users to delete their personal data is a legal requirement of the GDPR.

It it can't be enabled by default, I propose to add this at least to `sympa.conf` so admins are aware of it and can easily enable it.

Thanks, and keep up the good work!

ikedas:
@ilf , this PR has been successfully closed.  Please submit the new issue to discuss about more improvements.

PR #522, opened 2019-01-09
(review.tt2) Make column header for "delete" checkboxes visible in owner/moderator edit panel

mpkut:
When editing owners/moderators, the column for the "delete" checkboxes does not display a visible header like the other columns do.

This commit uses a simple ``<label>`` tag rather than an empty ``<i>``, which seems to be used elsewhere for hover-over labels.

### DIFF
--- a/default/web_tt2/review.tt2
+++ b/default/web_tt2/review.tt2
@@ -279,8 +279,8 @@
 
   [% IF pS.privilege == 'write' && is_privileged_owner ~%]
     <div class="small-2 medium-1 columns" role="columnheader">
-      <label title="[%|loc%]Delete[%END%]">
-        <i class="fa fa-user-times"></i>
+      <label>
+         [%|loc%]delete[%END%]
       </label>
     </div>
   [%~ END %]


mpkut:
Here is a screenshot of the owner edit panel demonstrating the empty header for the "delete" column:

owner-edit-before

And here is a screenshot with the update applied, showing the localized word "delete": 

owner-edit-after

ikedas:
Hi @mpkut, couldn't you see an icon and tooltip?
pr-522

What are the version and installation method of your Sympa?

mpkut:
Hmm. We're using 6.2.38, installed using RPM packaging drawn from your repo, however under RHEL 7 instead of Fedora. We are missing an icon font somewhere then. With your agreement, I will withdraw this PR and address it from the packaging side.

ikedas:
@mpkut, I agree to move to [ikedas/sympa-rpm](https://github.com/ikedas/sympa-rpm).

mpkut:
Yep, definitely packaging. An RPM build with ``unbundle_fontawesome`` disabled for RHEL 7 now shows the ``fa-user-times`` icon.  Closing this PR - apologies for the ignorance of just what we were looking at.

PR #528, opened 2019-01-15
Invalid default scenarios

ikedas:
Fixed bugs:

  - WWSympa: On list config page, default settings of list scenarios are ignored. And if user updated configuration, the first option in select box is saved.
    Fixed by assigning defaults in Sympa::ListDef.
  - `archive.web_access` list parameter did not have default value.

Change:
  - The feature making a symbolic link "*.default" to assign default values of scenarios is deprecated.
    It was hard to determine original option value linked to default.

### DIFF
--- a/default/Makefile.am
+++ b/default/Makefile.am
@@ -410,23 +410,6 @@ EXTRA_DIST = $(nobase_default_DATA) \
 
 CLEANFILES = list_aliases.tt2
 
-DEFAULT_SCENARIOS = \
-	add.owner \
-	archive_mail_access.closed \
-	d_edit.owner \
-	d_read.private \
-	del.owner \
-	info.open \
-	invite.private \
-	remind.owner \
-	review.owner \
-	send.private \
-	subscribe.open \
-	topics_visibility.noconceal \
-	tracking.owner \
-	unsubscribe.open \
-	visibility.conceal
-
 list_aliases.tt2: Makefile list_aliases.tt2.in
 	@rm -f $@
 	$(AM_V_GEN)$(SED) \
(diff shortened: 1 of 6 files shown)


dverdin:
cool!
I was about to argue about *.default but you're right: we don't care about it. It was never actually used.
So, yes, godspeed!

dverdin:
Hey, I'm stupid.
I can't believe I did not see it: how do you define the default scenario now?
Except by modifying ListDef, of course; I mean: how do you define the default scenario when you're an administrator? Most of these parameters don't have a sympa.conf default value.
An people having defined custom default scenarios will have their defaults ignored when upgrading.
Soji, we fucked up. This PR is a disaster. Please roll back and let's think  about it again.

ikedas:
Except by modifying ListDef? None. As I summed up, the feature defining defaults by symblic link haven't worked correctly. In other words, there have never been such ways.

dverdin:
Wait... So this feature NEVER worked?

dverdin:
Then, how does Sympa determines the default scenario for now?

ikedas:
@dverdin, are you feeling calm now?

I added robot-/site- level defaults to list scenario parameters, as the other list parameters have.  See #540.
If you found something wrong, please let us know.

I'll merge it in these weekdays.

dverdin:
@ikedas : Yes I do. Sorry for losing my calm. I should know better and trust your judgement since the time...
Setting the defaults in the server / virtual host configuration is perfect to me. 
So you got rid of a confusing bug and even improved Sympa.
As far as I'm concerned, it's OK to close these issues.
And in the future, I'll take deep breathes before commenting.
Regards,
David

Issue #527, opened 2019-01-12
Public archives not available with sympa-6.2.38


pascalmaes:
A user not logged in can't see the archive of a list with a public web access 

`ERROR (confirm_action) - Authorization rejected. Maybe you forgot to log in? `

Version
-------
6.2.38

Installation method
-------------------
From source on a virgin system (the previous installation of sympa has been deleted).
Everything as default.

Expected behavior
-----------------
The archive should be readable

Actual behavior
---------------

`ERROR (confirm_action) - Authorization rejected. Maybe you forgot to log in? `

In sympa.log

`Jan 12 11:20:04 sympa-test wwsympa[30577]: info main::do_arc(2019-01, /) [robot listes] [session 95681996162673] [client y] [list test-4]`
`Jan 12 11:20:04 sympa-test wwsympa[30577]: info main::check_action_parameters() [robot listes] [session 95681996162673] [client y] [list test-4] CSRF token mismatch: in="" session="aaad607574282c8ea3a865514c3ba433"`

In sympa.err

`Jan 12 11:20:04 sympa-test wwsympa[30577]: err main::#1557 [robot listes] [session 95681996162673] [client y] Missing required parameters for action "confirm_action"`

Additional information
----------------------
When logged as listmaster, I can't see the scenario source of "access right (web_access)"

` INTERNAL SERVER ERROR (dump_scenario) - `

In sympa.err 

`Jan 12 11:15:32 sympa-test.sipr.ucl.ac.be wwsympa[30577]: err main::#1572 > main::do_dump_scenario#15942 > Sympa::Scenario::new#77 Missing parameter`

ikedas:
Hi @pascalmaes,

Could you please tell us these things?
  - detailed operation you performed (launched browser, went to URL xxx, clicked xxx, ...).
  - web browser(s) you tried.

Thanks.

pascalmaes:
I have try with Safari, Firefox, Google Chrome on macOS and with Firefox on Windows

I know the URL : https://sympa-test.sipr.ucl.ac.be/sympa/arc/test-4/2019-01/

ikedas:
@pascalmaes, I think it's a bug.

Can you do either one of below?

  - Add this line to `sympa.conf` (after doing this, service have to be reloaded)
    ```
    web_archive_spam_protection javascript
    ```
  - Change config of corresponding list to set "Archives" / "email address protection method" to "javascript".

pascalmaes:
[...]

PR #529, opened 2019-01-15
Remove 'confirm_action' from list of actions requiring the CSRF token

mpkut:
The current CSRF code is configured to require the token for the `confirm_action` action. However #527 shows that there are legitimate Sympa functions such as unauthenticated archive views that invoke `confirm_action` as part of a GET request (i.e. the "I am not a spammer" click-through challenge for archives). 

This simple patch removes `confirm_action` from the `%require_csrftoken` list. That allows `confirm_action` to be invoked during GET requests, allowing unauthenticated archive views and other GET based confirmed actions to work as expected.

Thankfully, the input validation in `Sympa::WWW::Session::confirm_action()`  acts against CSRF on its own, so we are still covered there. The `confirm` submission must be the very next click in the user's Sympa session, otherwise the pending action is cancelled. Furthermore the action arguments must match a recorded hash before the action can proceed.

### DIFF
--- a/src/cgi/wwsympa.fcgi.in
+++ b/src/cgi/wwsympa.fcgi.in
@@ -730,7 +730,6 @@ our %required_privileges = (
 
 our %require_csrftoken = (
     'add'            => 1,
-    'confirm_action' => 1,
     'del'            => 1,
     'move_user'      => 1,
     'savefile'       => 1,


ikedas:
@mpkut, thanks for follow up.

Merged to a tentative branch [`sympa-6.2.40`](https://github.com/sympa-community/sympa/tree/sympa-6.2.40) for the next patch release.

PR #534, opened 2019-01-25
Mail command unavailable in confirmation requests

ikedas:
Even if web interface was disabled (wwsympa_url parameter was unset), mail URL link was not shown in confirmation requests for subscribe/unsubscribe actions.
Fixed by replacing test on an unassigned variable user_interfaces with test on conf.wwsympa_url.

### DIFF
--- a/default/mail_tt2/request_auth.tt2
+++ b/default/mail_tt2/request_auth.tt2
@@ -1,10 +1,8 @@
 [%# request_auth.tt2 ~%]
 To: [% to %]
-[% IF user_interfaces.size() == 1 and user_interfaces.0 == 'mail' -%]
-Subject: [% FILTER qencode %]AUTH [%keyauth%] [%cmd%][%END%]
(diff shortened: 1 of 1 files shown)


ikedas:
Background: Since 6.2.36, web interface can be enabled/disabled by each domain.  If it is enabled, web URL should be used. Otherwise, mail link must be used.

racke:
That makes sense.

dverdin:
> Background: Since 6.2.36, web interface can be enabled/disabled by each domain. If it is enabled, web URL should be used. Otherwise, mail link must be used.

Oh, nice. I had not noticed this feature.
Side question: do you disable bot web and SOAP or can you distinguish between both interfaces?

ikedas:
> Side question: do you disable bot web and SOAP or can you distinguish between both interfaces?

They may be disabled/enabled separately.  See also #330.

dverdin:
Cool, thanks!

PR #536, opened 2019-01-25
Revert "Invalid default scenarios"

dverdin:
Reverts sympa-community/sympa#528
A lot of problems with this PR. Sorry I did not see them before.
The problem is: listmasters can't define default scenarios anymore. the only way to do so is in ListDef.pm
Consequently : all instances having defined default scenarios will have their customizations ignored when upgrading.
This PR is clearly a regression and should be removed.

### DIFF
--- a/default/Makefile.am
+++ b/default/Makefile.am
@@ -410,6 +410,23 @@ EXTRA_DIST = $(nobase_default_DATA) \
 
 CLEANFILES = list_aliases.tt2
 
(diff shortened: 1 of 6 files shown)


ikedas:
Please don't submit simjple revert. Instead you would be better to discuss on the problem.

dverdin:
OK, sorry.

dverdin:
I did, but I thought it would be better to start a revert.

PR #540, opened 2019-01-28
Invalid default scenarios: Follow up to #528

ikedas:
Please comment on #528.


dverdin:
Very good complement to #528 . I agree.

PR #541, opened 2019-01-29
Owner page is empty

ikedas:
[bug] "Users" - "Owner" page is empty for a while after restart, as picture below.

empty_review_owner

Fixed by correcting typos not loading information soon.


dverdin:
Never noticed this.
However, thanks for fixing it @ikedas !

Issue #546, opened 2019-02-01
Exclusion robot could not be guessed for user


bertaviader:
We are migrating our sympa to another server and with a new version. I'm doing the upgrade of the database to the new version and this error appears:

err main::#744 > Sympa::Upgrade::upgrade#883 Exclusion robot could not be guessed for user "X" in list "Y". Either this user is no longer subscribed to the list or the list appears in more than one robot (or the query to the database failed). Here is the list of robots in which this list name appears: ""

In this case we have two scenarios, either the user is subscribed and the error still appears or the user is subscribed and in the same time in the exclusion table of the interface. 

Version
-------
The previous version is Sympa 6.1.7
The new version is Sympa 6.2.24

Installation method
-------------------
apt-get 

Expected behavior
-----------------
The user is not subscribed and inside the table of exclusion of the interface.

Actual behavior
---------------
The user is subscribed and the error still appears when I do the upgrade or the user is subscribed and in the same time inside the exclusion table of the interface.

bertaviader:
Hi,
The result of the query is:
```
+-----------------+------------------+
| robot_exclusion | family_exclusion |
+-----------------+------------------+
|                 |                  |
+-----------------+------------------+
```
I guess this is not a normal result, can you guide me to the proper solution?
Thank you!

ikedas:
It's normal result.  The next, what is the result of this query?
```
SELECT COUNT(*) FROM subscriber_table WHERE list_subscriber = 'Y' AND user_subscriber = 'X';
```
If the user was really subscribing, the result would be a row containing `1` or more.

bertaviader:
The result is this one:
```
+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
```
But in the web portal I can see the user subscribed to the list

dverdin:
Hi @bertaviader !
Just to avoid the obvious: you did replace 'Y' by an actual list name and ''X' by a user address in @ikedas queries, didn't you?
Because you can't see subscribed people in the web interface if they are not in the database.
Other possibility: Sympa is querying your former database. Check the "db_*" parameters in your sympa.conf

bertaviader:
Sorry! It was a mistake with the domain in the case above. 
But this problem persist: the user is subscribed and in the same time in the exclusion table of the interface.
In this case the first query return this
```
+-----------------+------------------+
| robot_exclusion | family_exclusion |
+-----------------+------------------+
| llistes.upc.edu |                  |
[...]

PR #549, opened 2019-02-07
exclusion_table might not be updated properly during "sympa.pl --upgrade"

ikedas:
Upgrading from 6.1.11 or earlier, `exclusion_table` might not be updated properly.

Fixed by updating the table ---
  1. not according to list membership of user in user_exclusion field.
     Currently the lists in which the user is not subscribing are skipped.
  2. removing the rows with the name of list which does no longer exist.
     Currently Sympa reports that _zero_ candidates for robot are there.

See also #546.


racke:
So this would not affect updates from Debian stretch (6.2.16) ?

ikedas:
See the change this pr will make.  It will affect update from 6.1.11 or earlier.

PR #567, opened 2019-03-11
Refactoring on mail templates

ikedas:
Simplify notation of header fields in mail templates.
  - Use "`[% "string" | loc %]`" instead of "`[%|loc%]string[%END%]`".
  - Use a new filter "`mailbox`" to construct address field in header.


dverdin:
Nice, Soiji. So Sympa supports both notation now, but the simpler is the default one, is it?

PR #583, opened 2019-04-03
Slight change in list admins cache expiry.

dverdin:
When changing a list's admin two times in the same second, the list admins cache is not updated. Consequently, when trying to add a third admin in the same second, the second one is not in the cache at the time.

Obviously it does not sound like a recurrent scenario and I met this problem when running a test file only.
However, it is possible that this happens more often when trying to make batch modifications in the shell. And also, it made my test fail which is an engineering problem: it could be misguiding about the cause of a test failure.
So I changed the expiration test, replacing "<" by "<=" and it solves the problem.
I don't see any side problem that could result from this change bu I commit it separately anyway because the cache is very sensitive.

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -693,7 +693,7 @@ sub _cache_get {
     $self->{_mtime}{$type} = $mtime;
 
     return undef unless defined $lasttime and defined $mtime;
-    return undef if $lasttime < $mtime;
+    return undef if $lasttime <= $mtime;
     return $self->{_cached}{$type};
 }
 


ikedas:
I feel it's good.
In the future, on-memory cache would be better to be removed.

dverdin:
Oh really? I thought it was introduced to fix performances problems.
Is it not good enough for this? Do you think there is a better solution?

racke:
The cache caused tons of problems.

dverdin:
Indeed but I thought it was fixed by now. But then again, I may have missed some conversations.

racke:
We used up a whole package of band aids, but the patient haven't recovered yet :roll_eyes:. Caching incorrectly causes more harm than it helps, especially home grown solutions like we got introduced into Sympa.

dverdin:
That should be a dedicated issue.
So I created one about open discussion regarding cache in Sympa: #584.
The current PR is really a light change in current cache.

PR #586, opened 2019-04-05
Fix #548 — Add DB log entry when message is rejected or distributed via email

### LINKED ISSUE #548: No log entry when message is rejected via email
Hello,

I've tested a few lists, with a few users, and consistently hitting this issue of a missing log entry. I'm unsure if it's my installed that's borked or an actual Sympa bug.

Version
-------
Sympa version 6.2.40 on RHEL 7.6 with postfix 2.10.1

Installation method
-------------------
Installed via Sympa-JA.org rpm

Expected behavior
-----------------
A log entry visible on the list page logging when a message was rejected by a moderator via email. The rejection is correctly logged if done via the web interface.

Actual behavior
---------------
No visible log entry on the list page

Thanks for looking into it!

Cheers

 Chris

ldidry:


dverdin:
That's a simple log addition, so it's quite straightforward to me:
- you respect the database fields constraints,
- you use the correct variables.
It looks ok to me.
A LITTLE CHANGE HOWEVER: On my opinion, the target_email should be the original author of the message. It's easy, you have it in `$message->{sender}`
I made a comment under the concerned lines in your commits.

dverdin:
(I don't see my comments, so I add it here:)

Replace:
`'target_email' => $param->{rejected_by},`
with:
'target_email' => $message->{sender},

src/lib/Sympa/Spindle/ProcessModeration.pm
```diff
@@ -151,6 +151,15 @@ sub _reject {
             },
             date => time + 1
         );
+        $log->db_log(
+            'robot'        => $list->{'domain'},
+            'list'         => $list->{'name'},
+            'action'       => 'reject',
+            'parameters'   => $message->{authkey},
+            'target_email' => $param->{rejected_by},
```
dverdin: Replace:
'target_email' => $param->{rejected_by},
with:
'target_email' => $message->{sender},

src/lib/Sympa/Spindle/ProcessModeration.pm
```diff
@@ -195,6 +204,16 @@ sub _distribute {
         Sympa::send_dsn($message->{context}, $message, {}, '2.1.5');
     }
 
+    $log->db_log(
+        'robot'        => $list->{'domain'},
+        'list'         => $list->{'name'},
+        'action'       => 'distribute',
+        'parameters'   => $message->{authkey},
+        'target_email' => $self->{distributed_by},
```
dverdin: Replace:
'target_email' => $param->{rejected_by},
with:
'target_email' => $message->{sender},

ldidry:
> A LITTLE CHANGE HOWEVER: On my opinion, the target_email should be the original author of the message. It's easy, you have it in $message->{sender}

I just copied the behavior of the logging of web reject/distribute but I agree that it makes more sense.

I changed it according to your comments :slightly_smiling_face:

dverdin:
Great!
The rest of the PR is perfect to me.
Have a nice week-end!

src/lib/Sympa/Spindle/ProcessModeration.pm
dverdin: Modification confirmed.

src/lib/Sympa/Spindle/ProcessModeration.pm
dverdin: Modification confirmed.

PR #582, opened 2019-04-01
Fix tt2 indentation

ldidry:
Files to review (I already reviewed them, but more reviewers is better) (remember to hide whitespace changes before reviewing, `git diff -w` on the CLI): 
- [ ] default/web_tt2/active_lists.tt2
- [x] default/web_tt2/admin.tt2
- [ ] default/web_tt2/ajax.tt2
- [ ] default/web_tt2/arc.tt2
- [ ] default/web_tt2/arc_manage.tt2
- [x] default/web_tt2/arcsearch.tt2
- [ ] default/web_tt2/arcsearch_form.tt2
- [ ] default/web_tt2/aside_menu.tt2
- [ ] default/web_tt2/blacklist.tt2
- [x] default/web_tt2/button_footer.tt2
- [x] default/web_tt2/button_header.tt2
- [ ] default/web_tt2/choosepasswd.tt2
- [ ] default/web_tt2/compose_mail.tt2
- [x] default/web_tt2/config_common.tt2
- [x] default/web_tt2/confirm_action.tt2
- [ ] default/web_tt2/copy_template.tt2
- [x] default/web_tt2/crash.tt2
- [x] default/web_tt2/create_automatic_list_request.tt2
- [ ] default/web_tt2/create_list_request.tt2
- [x] default/web_tt2/css.tt2
- [x] default/web_tt2/css_ie.tt2
- [x] default/web_tt2/d_control.tt2
- [x] default/web_tt2/d_editfile.tt2
- [x] default/web_tt2/d_install_shared.tt2
- [x] default/web_tt2/d_properties.tt2
- [x] default/web_tt2/d_read.tt2
- [ ] default/web_tt2/d_upload.tt2
- [ ] default/web_tt2/docindex.tt2
- [ ] default/web_tt2/dump_scenario.tt2
- [ ] default/web_tt2/dumpvars.tt2
- [ ] default/web_tt2/edit.tt2
- [x] default/web_tt2/edit_attributes.tt2
- [x] default/web_tt2/edit_config.tt2
- [x] default/web_tt2/edit_list_request.tt2
- [ ] default/web_tt2/edit_template.tt2
- [ ] default/web_tt2/editfile.tt2
- [x] default/web_tt2/editsubscriber.tt2
- [ ] default/web_tt2/error.tt2
- [x] default/web_tt2/family_signoff_request.tt2
- [ ] default/web_tt2/footer.tt2
- [ ] default/web_tt2/get_biggest_lists.tt2
[...]

PR #592, opened 2019-04-10
PR#582 addition 1

ikedas:
Additional fix to PR #582 (1) sso_login.tt2: Don't include HTML elements in translated strings.

Note: Message catalog should be tweaked.


dverdin:
Very good Soji. We got these kind of HTML tags in the catalogues for too long.

ldidry:
Could you have a look at `ls_templates.tt2` too? There is another translated string containing HTML tags (and uglier tha in `sso_login.tt2`).

ldidry:
It's at the beginning of the file, line 8, you can’t miss it.

PR #593, opened 2019-04-11
Sympa::Aliases::Template creates exclusive lock on local storage

ikedas:
This behavior is useless when aliases file is shared by multiple hosts with network drive.
Fixed by creating lock at the same place as aliases file.


racke:
Makes sense to me :+1:

dverdin:
I don't see any side effect that could come from this.
It's good to me. :+1:

Issue #524, opened 2019-01-10
Owner and editors edition through the command line


dverdin:
Expected Behavior
-----------------
A lot of administrators have built scripts to mass edit config files, especially for owners provisioning.

Current Behavior
----------------
These scripts are broken as of Sympa 6.2.34 due to the fact that editors and owners are now located in the database only. These are the only list parameters having such an exception.

Possible Solution
-----------------
I plan to add options to sympa.pl to manage list adminsitrators. That way, these scripts will keep on working, simply by replacing the sed or whatever command the admins were using by the relevant sympa.pl command.

These options would make usage of the new Sympa code and use the new primitives to add / replace / delete adminstrators.

Actually, this would even be an improvement to previous Sympa versions because such options would have been very usefull before, for example when somebody leaves an institution and her replacement needs to take over the leaver's list.

Here is the intended interface I want to provide:

sympa.pl --add-admin --list=<listname> --vhost=<domain> --email=<an@email> --role=<owner|editor> --visibility=<value> --profile=<privileged|normal> --reception_mode=<reception_mode> --gecos<gecos>

sympa.pl --delete-admin --list=<listname> --vhost=<domain> --email=<an@email> --role=<owner|editor>

sympa.pl --replace-admin --list=<listname> --vhost=<domain> --previous_admin=<an@email> --new_admin=<an.other@email>

In each case, the "last modified by" list parameter would contain "listmaster@domain.tld" email.

What do you think? I could do this easily and quickly using the current codebase.

Context
-------
I'm in contact with a lot of listmasters through the lists and a lot complained about the lost functionnality, while agreeing the vanishing of the cache problem was very cool.

racke:
Wouldn't it be better to include the owners from SQL/LDAP/... like we do for subscribers?

ikedas:
Providing Sympa::Request handler at first seems better.  It may be useful for the other components than command line (e.g. SOAP: See #526).

racke:
Yes, definitely. That's the way to go.

ikedas:
Hi @dverdin,

### Alternative solution

If you seek a quick solution, you'd be better to consider using dump files.

Previously administrators edited config file to change owners/moderators directly.  Instead, with current version, they may do
```
sympa.pl --dump_users --list LISTNAME --role owner,editor
(Edit owner.dump and/or editor.dump)
sympa.pl --restore_users --list LISTNAME --role owner,editor
```
[...]

PR #594, opened 2019-04-11
List admin deletion through command line

dverdin:
This is the follow-up to PR #551.

See issue #524 for discussion about the feature.

I implemented the _deletion_ of a list admin (owner or editor) through the command line.
The code is based on a Request::Handler called delete_list_admin.

You can delete a list admin using the following command:
```
sympa.pl --delete_list_admin \
    --email <email@example.com> \
    --role [owner|editor]
    --list <listname@lists.domain.tld> \
    --robot <lists.domain.tld> \
```
Details about the command:
  - `--email` is mandatory.
  - If you don't specify `--role`, both editor and owner roles will be removed.
  - You also need to specify either `--list` or `--robot`.
    - `--list` is a full list id (<local.part>@<domain.part>) ; if you specify it, the selected role will be deleted for the user in this list only ;
    - `--robot` is the name of a Sympa robot ; if you specify it, the selected role will be deleted for the user in all this robot's lists.

For now, the command refuses to complete for owners only if the user to be deleted if the last list owner. This behaviour should probably be refined.
Here are my thoughts:
  - We could switch the last user with listmaster.
  - We could also add a "force" option that would ignore this protection.
  - It could also be a mitigation of both solutions:
    - for example, force the deletion but replace the email with the sender's email.

To be discussed, I guess.

Todo after it: implement list admin update (for which you could also specify the whole virtual host).


racke:
If you are fixing bugs in sympa.pl, that should clearly go into a **separate** PR.  And if you concatenate strings please add whitespace for readability.

dverdin:
Oh no, it's a bug in sympa.pl _in the part I just added_.
It is in no way a pre-existing bug.

dverdin:
(Whitespace added) :wink:

ikedas:
> * `--email` is mandatory.
> * If you don't specify `--role`, both editor and owner roles will be removed.
> * You also need to specify either `--list` or `--robot`.

IMO I prefer to make `--role` also mandatory.  As breaking adminship can stop the list, user would be better to know what they are doing.

racke:
> Oh no, it's a bug in sympa.pl _in the part I just added_.
> It is in no way a pre-existing bug.

Good, the commit message was not precise enough to tell me that.

dverdin:
Racke : yes, I'll make commit messages more explicit.

dverdin:
@ikedas : I agree. The reason why I did not make role mandatory was to ease deprovisionning, when you want to remove an address from a whole domain.
However, I added a control to prevent removing the last owner of a list.

ikedas:
Same as #551. Though I haven't investigated, how this will work when `owner_domain` feature is enabled?

racke:
Same problem as #551

dverdin:
Ah, OK, this came out of my sonar.
I'll address this problem.

racke:
Please:

- Always make sure to start from latest commit in sympa-6.2
- Never merge anything, rebase should be fine
- Always check commits you are pushing 
- Always review your PR (commits, files changed)

Enjoy!

dverdin:
No problem.
I'll clean this mess and create new, clear PR.
Thanks for your feedback and a happy new year to you!

PR #615, opened 2019-05-16
Prevent unnecessary DB access in add_list_member

seblgr:
In add_list_member the value $current_list_members_count is discarded if max_list_member == 0 (default setting).
This prevents the "high db load" produced by the successive DB accesses.


ikedas:
Merged. Thanks for improving Sympa!

PR #616, opened 2019-05-17
Add a button for full export of subscribers

ldidry:
- Currently, only light export is available, thus export doesn’t contain
  the name of the subscribers. The button solves that.
- The full export always includes a header:
  \# Exported subscribers with search filter ""
  But it’s useless and maybe confusing when you use the full export from
  the button. I removed that header if no filter was sent
- I changed an "unless defined" assignment to "//", I don't know if it’s
  ok to do that.

### DIFF
--- a/default/web_tt2/review.tt2
+++ b/default/web_tt2/review.tt2
@@ -46,6 +46,9 @@
         <a class="actionMenuLinks" href="[% 'export_member' | url_rel([list,'light']) %]">
             [%|loc%]Dump[%END%]
         </a>
+        <a class="actionMenuLinks" href="[% 'export_member' | url_rel([list]) %]">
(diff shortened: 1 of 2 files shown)


ikedas:
> * I changed an "unless defined" assignment to "//", I don't know if it’s
>   ok to do that.

Unfortunately Sympa has to support Perl 5.8.1 or later.  ``//`` appeared on Perl 5.10.

ikedas:
See #620.

ldidry:
Thanks for the hint and the new issue, I feared something like that. I’ll revert my `//` change and redo it when Perl 5.8 support will be officially dropped.

ldidry:
`//` removed.

PR #617, opened 2019-05-17
Fix #601 — Add a configure option for sympa.conf-dist location

### LINKED ISSUE #601: Add a configure option for sympa.conf-dist location
Related to #595, see https://github.com/sympa-community/sympa/pull/595#discussion_r277600734

ldidry:
I’m not sure about the `confdistdir` name of the option, feel free to propose something else.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -137,11 +137,11 @@ installconfig: installdir
 		chown $(USER) $(DESTDIR)$(confdir)/sympa.conf; \
 		chgrp $(GROUP) $(DESTDIR)$(confdir)/sympa.conf; \
 	fi
-	@echo "Installing configuration template ..."
-	-@echo "installing sympa.conf-dist"; \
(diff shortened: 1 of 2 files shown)


ikedas:
I won't oppose to add this new option, if anyone agree to it.

xavierba:
As I wrote on the mailing list when we've discussed this, I think installing sympa.conf-dist to sysconfdir is not correct and the file is going to be moved to docdir in rpm. @racke wrote the file will need to be moved to docdir too for deb.
Adding a configure option is slightly better than to have to move the file manually, but is that really necessary ? My take on this would be this is a doc file, install it in docdir and be done with it.

ldidry:
First, it’s not unusual nowadays to have a conf template next to the real conf. Then, not everybody looks at the doc directory so using confdir by default will help newcomers to find it.

ikedas:
I feel discussion might be continued at #601 (current page is on specific pr).

I know that some distributions prepare config templates on the same place as editable config for some software packages.  One issue is, I think, by whom they should be prepared: Source maintainer or each distributor.

PR #618, opened 2019-05-17
Fix #613 — Avoid MHonArc’s search warning flood

### LINKED ISSUE #613: Spurious MHonArc Search warnings for undefined search parameters
Version
-------
Version 6.2.42

Installation method
-------------------
Installed from Source on a Debian 9 System

Expected behavior
-----------------
No warnings when MHonArc archives are searched.

Actual behavior
---------------
Many warnings populate error log when archives are searched for undefined parameters: $subj, $date, $from, $id.

When someone does archive searching, our log files fill up with warnings such as:

```
[Tue May 14 09:19:44.283239 2019] [fcgid:warn] [pid 23730] [client <ipaddress>:4878] mod_fcgid: stderr: [Tue May 14 09:19:44 2019] wwsympa.fcgi: Use of uninitialized value $subj in substitution (s///) at /usr/local/sympa/bin/Sympa/WWW/Marc/Search.pm line 307., referer: <website>/sympa
```

Additional information
----------------------

I made some changes to <home>/sympa/bin/Sympa/WWW/Marc/Search.pm that seems to solve the issue, but was not confident to submit a patch.

```
--- Search.pm   2019-05-14 10:36:27.205977640 -0400
+++ Search.pm_dist      2019-05-14 09:12:25.280414786 -0400
@@ -304,20 +304,18 @@
                 last if (/-->/);
             }
         }
+        $subj =~ s/ *-->$//;
 
-        if($subj) { $subj =~ s/ *-->$//; }
-
-               ($from = (<$fh> // '') ) =~ s/^<!--X-From-R13: (.*) -->/$1/;
-
+        ($from = <$fh>) =~ s/^<!--X-From-R13: (.*) -->/$1/;
 
         ## No more need to decode header fields
         #$from = &MIME::Words::decode_mimewords($from);
 
         $from =~ tr/N-Z[@A-Mn-za-m/@A-Z[a-z/;
 
-               ($date = ( <$fh> // '') ) =~ s/^<!--X-Date: (.*) -->/$1/;
+        ($date = <$fh>) =~ s/^<!--X-Date: (.*) -->/$1/;
 
-               ($id = (<$fh> // '') ) =~ s/^<!--X-Message-Id: (.*) -->/$1/;
+        ($id = <$fh>) =~ s/^<!--X-Message-Id: (.*) -->/$1/;
 
         if ($body) {
             my $lines = '';
```

ldidry:
I think it’s a good fix for #613

### DIFF
--- a/src/lib/Sympa/WWW/Marc/Search.pm
+++ b/src/lib/Sympa/WWW/Marc/Search.pm
@@ -304,6 +304,11 @@ sub search {
                 last if (/-->/);
             }
         }
+
+        # If $subj is undefined, <$fh> will be undefined thus going further
+        # is useless
+        next unless defined $subj;
+
         $subj =~ s/ *-->$//;
 
         ($from = <$fh>) =~ s/^<!--X-From-R13: (.*) -->/$1/;


ldidry:
Wait until @pkissman confirmed it’s OK.

src/lib/Sympa/WWW/Marc/Search.pm
```diff
@@ -304,6 +304,11 @@ sub search {
                 last if (/-->/);
             }
         }
+
+        # If $subj is undefined, <$fh> will be undefined thus going further
+        # is useless
+        next unless defined $subj;
```
ikedas: I think it will miss messages matching with `$body` but not subject.

ldidry: If `$subj` is defined before the `while` loop: nothing changes.

If `$subj` is not defined before the `while` loop and is still not defined after that, `<$fh>` is no more defined, since it’s not a `last` that ended the `while` loop but the end of file.

`$lines` of line 326 will always be `''` since the `while` loop of line 327 will never start as `<$fh>` is not defined.

Don’t you think?

ldidry: Up

pkissman:
Thank you.

src/lib/Sympa/WWW/Marc/Search.pm
ikedas: I was confused. There look no problem.

PR #627, opened 2019-05-24
Fix #621 — Add a 'gecos' option to web_archive_spam_protection

### LINKED ISSUE #621: Hide full email addresses in archives
Expected Behavior
-----------------
Add possibility to hide email addresses in order to make archive public without exposing these sensitive informations.

Current Behavior
----------------
Currently, email addresses (and gecos) are displayed.

Possible Solution
-----------------
Add possibility to only display gecos?

Context
-------
You want to make discussion public but don't want to expose addresses.

ldidry:
This allows to make web archives public without fearing to leak users
email addresses. This has only effect when you display a mail. In the archive list, the behavior is the current one: display gecos if any, display first part of email address otherwise (i.e. `foo` if the address is `foo@bar.org`).

Please note that I adapted the code from `decorate_email_at` subroutine.

Also, I used 'No gecos' as message when the email address didn’t contain a gecos, but feel free to propose somethin more easier to understand.

May fix #621

### DIFF
--- a/src/lib/Sympa/HTMLDecorator.pm
+++ b/src/lib/Sympa/HTMLDecorator.pm
@@ -27,6 +27,7 @@ package Sympa::HTMLDecorator;
 use strict;
 use warnings;
 
+use Sympa::Language;
 use Sympa::Regexps;
 use Sympa::Tools::Text;
 
@@ -184,6 +185,7 @@ sub decorate {
     if ($options{email}) {
         $self->{_shdEmailFunc} =
               $options{email} eq 'at'         ? \&decorate_email_at
+            : $options{email} eq 'gecos'      ? \&decorate_email_gecos
             : $options{email} eq 'javascript' ? \&decorate_email_js
             :                                   undef;
     }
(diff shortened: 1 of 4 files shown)


src/lib/Sympa/HTMLDecorator.pm
```diff
@@ -230,6 +232,30 @@ sub decorate_email_at {
...
+sub decorate_email_gecos {
+    my $self = shift;
+
+    my $decorated = '';
+    my $email_re  = Sympa::Regexps::addrspec();
+    my $language  = Sympa::Language->instance;
+    while (my $item = $self->_queue_shift) {
+        if ($item->{event} eq 'text') {
+            my $dtext = Sympa::Tools::Text::decode_html($item->{text});
+            if ($dtext =~ s{<?\b($email_re)\b>?}{}g) {
```
ikedas:
```
            if ($dtext =~ s{<?\b($email_re)\b>?}{}g) {
```
E-mail and similar texts will entirely disappear.  Code such as
```
            my $replacement = $language->gettext('address@concealed');
            if ($dtext =~ s{\b$email_re\b}{$replacement}g) {
```
may be better.

ldidry: :+1: 
I will update my PR this week

ldidry: Done.

ikedas: I confirmed.

Issue #581, opened 2019-03-31
Extra space at the end of line in topics.conf


ikedas:
Version
-------

6.2.40

Installation method
-------------------

RPM (Sympa-JA.org)

Actual behavior
---------------

If `topics.conf` has this content:
``` code
other
visibility conceal␣
```
(``␣`` is a space)

Each time the users access to web interface, this error message is shown in the log:
``` code
Unable to find scenario file "topics_visibility.conceal "
```

dverdin:
To fix this, we might change Conf.pm to ignore trailing spaces.

racke:
Yeah good idea. As a rule, always strip surrounding spaces.

dverdin:
Damn. We seem to be doing it already. It was too easy...
https://github.com/sympa-community/sympa/blob/2766e0ef8a536e9726757865cfc8e9050738edcf/src/lib/Conf.pm#L1737

PR #629, opened 2019-05-24
Don’t capture extra spaces at the end of line in topics.conf

ldidry:
I didn’t have the same error described in #581 (`Unable to find scenario file "topics_visibility.conceal "`) but this one:

```
DIED: Can't call method "authz" on an undefined value at /home/sympa/bin/wwsympa.fcgi line 15627.
 at /home/sympa/bin/wwsympa.fcgi line 15627.
	main::export_topics('sympa.test.org') called at /home/sympa/bin/wwsympa.fcgi line 2755
	main::check_param_in() called at /home/sympa/bin/wwsympa.fcgi line 1501
```

My patch fixes it, so it should work for #581

### DIFF
--- a/src/lib/Sympa/Robot.pm
+++ b/src/lib/Sympa/Robot.pm
@@ -227,7 +227,7 @@ sub load_topics {
                     'name'  => lc($1),
                     'order' => $index
                 };
-            } elsif ($line =~ /^([\w\.]+)\s+(.+)\s*$/) {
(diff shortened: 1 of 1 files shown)


ldidry:
Obviously, not the same error, but same cause of error (extra space in `topics.conf`)

ikedas:
Looks good for me 👍

PR #638, opened 2019-06-05
Make it clear on web interface that a list is being moderated

### LINKED ISSUE #636: Make it clear that a list is being moderated
Expected Behavior
-----------------
Users should be clearly informed that their list is being moderated. 

Current Behavior
----------------

Currently, there is a message in the left side menu that's said: **List not activated yet**. The message is not very visible.

Possible Solution
-----------------
Maybe an alert box that's said: **List not activated yet. Please wait until a listmaster activated it.**:
Sélection_141

Context
-------

When the list creation is moderate, users are not sufficiently informed that their list is not yet available and write to the listmaster to find out what is happening.

ldidry:
Fix #636

Adds an alert-box with the text "List not activated yet. Please wait until a listmaster activated it to be able to use it." below list subject.

### DIFF
--- a/default/web_tt2/main.tt2
+++ b/default/web_tt2/main.tt2
@@ -84,6 +84,11 @@
                             [% END %]
                         </p>
                     [% END %]
+                    [% IF list_status == 'pending' %]
+                        <p class="alert-box warning text-center">
+                            [%|loc%]List not activated yet. Please wait until a listmaster activated it to be able to use it.[%END%]
+                        </p>
+                    [% END %]
                 [% END %]
 
                 [% PROCESS nav.tt2 %]


racke:
Adding the alert box is a good idea, however I don't like the second sentence. We can't give a clue when the list is activated - maybe it never happens. What about "Please contact your listmaster for further questions."

ldidry:
I’m agnostic on that. @SansPseudoFix, what do you think?

SansPseudoFix:
I understand your concern but if we replace it by "Please contact your listmaster for further questions.", I'm afraid we will receive **a lot** of questions ("When do you think activated my list?"). Imho if people really have questions, they contact us.

ikedas:
I support @SansPseudoFix.

ikedas:
I think we'd be better to merge this pr and then think about better text.  Is that ok?

ldidry:
Sure: anything is better than nothing :+1:

racke:
OK fair enough.

ikedas:
Merged. Thanks for comments.

PR #644, opened 2019-06-07
Add trusted status to users, allowing to use "is_trusted" (new) in scenarii

### LINKED ISSUE #633: Is it possible to moderate list creation and after say "That user is legit and can bypass moderation"?
[Framalistes](https://framalistes) is a public Sympa instance. Anyone can create lists on it.

Spammers used it the last two weeks to create a lot of lists and send spam.

We now use the `create_list.public_listmaster` scenario to avoid that. It solves the spam issue but slows down the "create list and use it" process for legit users.

Is there a way to put new users in quarantine and make their list creation moderated while having a flag on manually verified users that make them able to create lists without being moderated?

If there is no way to do that, I’ll make this issue a feature request and work on it.

ldidry:
- Add "is_trusted" authorization to scenario create_list.public_listmaster
- You can change the trust status when activating pending list
- You can manage trust status through serveradmin/users, for one user or
  all trusted users
- You can see trusted users and manage trust status with sympa.pl

Fix #633

### DIFF
--- a/default/Makefile.am
+++ b/default/Makefile.am
@@ -311,6 +311,7 @@ nobase_default_DATA = \
 	web_tt2/maintenance.tt2 \
 	web_tt2/main.tt2 \
 	web_tt2/manage_template.tt2 \
+	web_tt2/manage_trusted_users.tt2 \
 	web_tt2/modindex.tt2 \
 	web_tt2/my.tt2 \
 	web_tt2/nav.tt2 \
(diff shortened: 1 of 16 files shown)


ikedas:
I think defining specific privileges may be realised by using [named filters](https://sympa-community.github.io/manual/customize/basics-scenarios.html#named-filters) in scenario.

ldidry:
Closing. Replaced by a custom scenario condition thanks to @ikedas suggestion on #633.
My custom scenario condition is waiting for merge on https://github.com/sympa-community/sympa-contribs/pull/10

PR #645, opened 2019-06-07
Allow to activate pending lists with sympa.pl

### LINKED ISSUE #637: Handle pending lists with sympa.pl
I wish I could see the pending lists with sympa.pl and be able to activate those lists with it too.

Proposition:
```
sympa.pl --show-pending-lists
sympa.pl --activate-lists list1,list2,list3
sympa.pl --activate-lists --all
```

ldidry:
- Small change to `--open_list` to let it handle pending lists
- Add `--show_pending_lists=robot` option to print pending lists (with some
  informations)

Fix #637

### DIFF
--- a/src/sbin/sympa.pl.in
+++ b/src/sbin/sympa.pl.in
@@ -87,7 +87,7 @@ unless (
         'send_digest',                  'keep_digest',
         'upgrade_config_location',      'role=s',
         'dump_users',                   'restore_users',
-        'open_list=s',
+        'open_list=s',                  'show_pending_lists=s'
     )
 ) {
     pod2usage(-exitval => 1, -output => \*STDERR);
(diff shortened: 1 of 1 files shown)


ldidry:
:thinking: It lacks a `--reject-list` option though. I’ll try to add it next week.

ldidry:
In fact, `close_list` can be used to reject pending list. No need for another option.

@ikedas, @racke, could you review this PR, please ?

ldidry:
> Should we modify close_list handler?

No.

Well, in fact, if we want to send a notification to the owner, we should (I don’t know if it would be hard). It should be a `--notify` option that could be used in combination with `--close_list` and `--open_list`. What do you think ?

ikedas:
> > Should we modify close_list handler?
> 
> No.
> 
> Well, in fact, if we want to send a notification to the owner, we should (I don’t know if it would be hard). It should be a `--notify` option that could be used in combination with `--close_list` and `--open_list`. What do you think ?

Oh sorry I see. You already have added `mode` attribute.
Currently `--open_list` does not have `--notify` option, and I think `--close_list` may also not have it.

I'll approve this pr and merge it later.

PR #675, opened 2019-06-28
Add ability to forbid some list names

### LINKED ISSUE #672: need ability to ban list names and owners
In operating a public sympa instance that allows anyone to create an account and request a list, we have had a problem with some users creating abusive lists. When we shutdown the list and close the owner account, they just create another, and this happens over and over. The cases we've seen haven't done this for general spam, but seem to usually be on a particular topic. At first they are honest when making the list request, but once we notice the pattern and stop approving those lists, they will lie in the list request and use other email addresses to submit the requests. But then after approval they will rename the list and often set the owner to an address used previously.

A while ago, to combat this problem, we implemented a nagios check that used regular expressions to scan the database and look for listnames and owner addresses that matched the people doing this and then alerted us to the problem. That works, but there wasn't anything automatic to close the list or account.

What would be best is if this could be prevented in sympa, at account and list request creation time and also when renaming lists.

ldidry:
- Add prohibited_listnames setting (comma-separated list, with ability
to use a wildcard)
- Add prohibited_listnames_regex setting

The settings are cumulative. You can use one or the other or both.

May fix #672

### DIFF
--- a/src/lib/Sympa/Aliases.pm
+++ b/src/lib/Sympa/Aliases.pm
@@ -135,6 +135,28 @@ sub check_new_listname {
             {new_listname => $listname});
     }
 
+    # Prevent to use prohibited listnames
+    my $regex = '';
+    if ($Conf::Conf{'prohibited_listnames_regex'}) {
+        $regex = lc($Conf::Conf{'prohibited_listnames_regex'});
+    }
+    if ($Conf::Conf{'prohibited_listnames'}) {
+        foreach my $l (split ',', lc($Conf::Conf{'prohibited_listnames'})) {
+            $l =~ s/\*/.*/g;
+            $l = sprintf('^%s$', $l);
+
+            if ($regex) {
+                $regex .= '|' . $l;
+            } else {
+                $regex .= $l;
+            }
+        }
+    }
+    if ($regex && $listname =~ m/$regex/) {
+        $log->syslog('err', 'Prohibited "%s"', $listname);
+        return ('user', 'prohibited_listname', {argument => $listname});
+    }
+
     # Check listname on SMTP server.
     my $aliases =
         Sympa::Aliases->new(Conf::get_robot_conf($robot_id, 'alias_manager'));
--- a/src/lib/Sympa/Aliases.pm
+++ b/src/lib/Sympa/Aliases.pm
@@ -138,11 +138,13 @@ sub check_new_listname {
     # Prevent to use prohibited listnames
     my $regex = '';
     if ($Conf::Conf{'prohibited_listnames_regex'}) {
-        $regex = lc($Conf::Conf{'prohibited_listnames_regex'});
+        $regex = eval(sprintf 'qr(%s)',
+            $Conf::Conf{'prohibited_listnames_regex'} // '');
     }
     if ($Conf::Conf{'prohibited_listnames'}) {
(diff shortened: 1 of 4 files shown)


racke:
Please get rid of the unnecessary white spaces in List.pm. Also I would like to make suggestions on the wording for messages and instructions.

ldidry:
The white spaces in List.pm comes from perltidy so, if we want to pass perltidy test (I launch it in a pre-commit hook, but it’s launched in CI too), we need them.

For the wording, please, go ahead, make suggestions :slightly_smiling_face:

racke:
I would still prefer to kill whitespace in a separate PR. This makes reviewing more difficult IMHO.

ldidry:
As you want :slightly_smiling_face:

ldidry:
Done (PR: #677)

src/lib/Sympa/Aliases.pm
```diff
@@ -135,6 +135,28 @@ sub check_new_listname {
...
     }
 
+    # Prevent to use prohibited listnames
+    my $regex = '';
+    if ($Conf::Conf{'prohibited_listnames_regex'}) {
+        $regex = lc($Conf::Conf{'prohibited_listnames_regex'});
+    }
+    if ($Conf::Conf{'prohibited_listnames'}) {
+        foreach my $l (split ',', lc($Conf::Conf{'prohibited_listnames'})) {
+            $l =~ s/\*/.*/g;
```
ikedas: Regex metacharacters should be escaped:
```perl
$l =~ s/([^\s\w\x80-\xff])/\\$1/g;
$l =~ s/\\\*/.*/g;
```

src/lib/Sympa/Aliases.pm
```diff
@@ -135,6 +135,28 @@ sub check_new_listname {
             {new_listname => $listname});
     }
 
+    # Prevent to use prohibited listnames
+    my $regex = '';
+    if ($Conf::Conf{'prohibited_listnames_regex'}) {
+        $regex = lc($Conf::Conf{'prohibited_listnames_regex'});
+    }
```
ikedas: - Malformed regexps should be filtered, or Sympa might crash, for example,
    ``` perl
    $regex = eval(sprintf 'qr(%s)', $Conf::Conf{'prohibited_listnames_regex'}) // '';
    ```
  - `lc()`-ing regexp looks not good idea: E.g. `\W`, `\S` might be broken.

ldidry: I’ll use insensitive regex instead

src/lib/Sympa/Aliases.pm
ldidry: Done

src/lib/Sympa/Aliases.pm
ldidry: Done

ikedas: Sorry,
``` perl
            eval(sprintf 'qr(%s)', $Conf::Conf{'prohibited_listnames_regex'} // '');
```

ikedas: ping @ldidry

ldidry: I’ll do it this week.

ldidry: Sorry for the delay, I forgot. It’s done.

PR #676, opened 2019-06-28
Use new Sympa logo

### LINKED ISSUE #665: New Sympa Logo
Expected Logo
-----------------

New Sympa logo is proposed.

[](https://www.sympa.org/logos/)
(Click the image above to see the other variants.)

Current Logo
----------------

This is Sympa logo currently used.

Possible Solution
-----------------

Replacing Sympa logo and icon in the source distribution with new ones.

Context
-------

ldidry:
I choose the red one because Sympa is mainly blue and the blue logo was not very visible.

Should we ship the other colors in Sympa so users could use them if they changed Sympa’s default colors?

May fix #665


ikedas:
Closed for now.  See #676 .

PR #677, opened 2019-06-28
Fix tidying of src/lib/Sympa/List.pm

ldidry:
Commit separated from #675


ldidry:
Needed to pass perltidy test.

PR #713, opened 2019-07-31
WWSympa: Use the label 'Apply modifications' instead of 'Update'

### LINKED ISSUE #698: 6.2.44 Cannot delete a moderator/owner
Version
-------
6.2.44

Installation method
-------------------
RPM

Expected behavior
-----------------
When checking the box and the clicking the delete icon, it should delete the moderator or owner.

Actual behavior
---------------
Nothing happens. Dev console tells me there is no event listener on the icon.

ikedas:
Use 'Apply modifications' instead of 'Update' to make sure the changes/deletions will be effective only if update is performed.  These functions are subject to the change: `edit_list`, `editsubscriber/review/set` and `edit`.

This may fix #698.

### DIFF
--- a/default/web_tt2/edit.tt2
+++ b/default/web_tt2/edit.tt2
@@ -52,7 +52,7 @@
                 <input type="hidden" name="submit" value="submit" />
                 <div class="columns">
                     <input class="MainMenuLinks" type="submit" name="action_edit"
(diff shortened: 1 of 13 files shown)


ldidry:
Sorry, I didn’t saw that you asked for a review. Looks good to me :+1:

PR #714, opened 2019-07-31
Trim fonts flavors to TTF/OTF only

xavierba:
Nowadays, TTF and OTF font formats are supported in all browsers:
https://caniuse.com/#search=ttf

Subsequently, there's no need to clutter the distribution with font flavors that won't be used anyway.
Removing all but the TTF/OTF flavors for Font-Awesome and Foundation-Icons saves 350+KB.

Please note this change is unlikely to break any browser, as the third font shipped with Sympa, Raleway, is only distributed in the OTF flavor (and was only shipped in TTF flavor before #191).

The third commit is a fix to correct Raleway font format in default css. This supposedly minor bug, as it went unnoticed  for more than a year, was introduced in #191.


ldidry:
Please, do not trim .woff and .woff2 formats since they are lighter than ttf/otf (in fact, [they are compressed ttf/otf](https://en.wikipedia.org/wiki/Web_Open_Font_Format#Specification)). Those formats are the future, not the past like eot (that, as for me, you can remove).

Woff is well supported by browsers: https://caniuse.com/#search=woff
Woff2 is supported by modern browsers: https://caniuse.com/#search=woff2

ikedas:
At present we will partially adopt this PR to keep woff/woff2 formats.  Close.

xavierba:
@ikedas , sorry not updating the PR myself.

PR #740, opened 2019-09-02
Pending lists not clearly shown as "pending" in the web GUI (#738)

### LINKED ISSUE #738: Pending lists not clearly shown as "pending" in the web GUI
When list creation is moderated, such lists should be clearly tagged as "pending" in the web GUI

Version
-------
6.2.36

Installation method
-------------------
sources

Expected behavior
-----------------
When a user creates a list and it remains pending, the web GUI should show that list as "pending" both in the "my" page and in the list's "info" page.

Actual behavior
---------------
In 6.2.36 the "my" page does not indicate that a list has a pending status. It can lead to confusion, because list aliases are not installed for pending lists.

The info page includes "List not activated yet" (see attached screenshot), but it should be highlighted. It used to be highlighted via the `bg_color_error` style, but this style definition has been removed in revision dab09c33c249c18f8aef53311ddd24bf3143d58e. I did not find a suitable style to use as a replacement.

image

ikedas:
This may fix #738

### DIFF
--- a/default/web_tt2/editsubscriber.tt2
+++ b/default/web_tt2/editsubscriber.tt2
@@ -96,7 +96,7 @@
         </p>
 
         [% IF current_subscriber.bounce %]
-            <h3 class="bg_color_error">[%|loc%]Bouncing address[%END%]</h3>
+            <h3 class="warning label">[%|loc%]Bouncing address[%END%]</h3>
 
             [% IF current_subscriber.bounce_address    %]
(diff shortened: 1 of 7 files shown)


salaun-urennes1:
I tried Sympa from your PR branch : that's much better :-)
Thank you for the quick fix.

ikedas:
@salaun-urennes1, thanks for cheking! I merged it.

PR #741, opened 2019-09-02
Fix errors 'Can't use an undefined value as an ARRAY reference' while running sympa.pl --modify_list

salaun-urennes1:
I get this error while running 
```
$ sympa.pl -d --modify_list testfamily  --input_file /usr/local/sympa/list_data/testlist1/instance.xml
Can't use an undefined value as an ARRAY reference at /usr/local/sympa/bin/sympa.pl line 931.
```

I fixed `sympa.pl` to ensure `$result->{'string_info'} `and `$result->{'string_info'}` are defined before trying go through a `ARRAYREF`.

### DIFF
--- a/src/sbin/sympa.pl.in
+++ b/src/sbin/sympa.pl.in
@@ -987,15 +987,19 @@ elsif ($main::options{'modify_list'}) {
         "\n************************************************************\n";
 
     unless (defined $result->{'ok'}) {
-        printf STDERR "\n%s\n", join("\n", @{$result->{'string_info'}});
         print STDERR "\nThe action has been stopped because of error :\n";
-        printf STDERR "\n%s\n", join("\n", @{$result->{'string_error'}});
+        if (defined $result->{'string_error'} && (ref($result->{'string_error'}) eq 'ARRAY')) {
+            printf STDERR "\n%s\n", join("\n", @{$result->{'string_error'}});
+	}
         exit 1;
     }
 
     close INFILE;
 
-    printf STDOUT "\n%s\n", join("\n", @{$result->{'string_info'}});
+    if (defined $result->{'string_info'} && (ref($result->{'string_info'}) eq 'ARRAY')) {
+         printf STDOUT "\n%s\n", join("\n", @{$result->{'string_info'}});
+    }
+
     exit 0;
 }
 


racke:
I would rather fix the place in the code which misses to put the information into string_error variable.

salaun-urennes1:
I think both make sense.
The best way to strengthen the code would be to encapsulate all the informations about function execution (`ok`, `string_info` and `string_error`) in a dedicated object. The object could have a print() or as_string() method that would take care of undefined cases.

racke:
Yeah indeed. I agree with you. What was the problem in your case?

ikedas:
@salaun-urennes1's changes will prevent crash by treating undefined value as arrayref.

FY, I'm planning to refactor --- or rewrite Sympa::Family so that it may not contain inconsistent error handling using hash elements.

salaun-urennes1:
The problem I'm facing: sympa.pl crashes.
Hopefully in the situation it's not too serious, because the job has been done by sympa.pl before it crashes. But it's confusing for the user and also prevents proper exit status handling in third-party scripts.

racke:
Can you share your XML file?

salaun-urennes1:
Here are the config files for the family initial instantiation :
```config.tt2
subject [% subject %]

[% FOREACH o = owner -%]
owner
email [% o.email %]
profile privileged

[% END -%]
```
```initial_definition.xml
 <family> 
   <list>
    <listname>testlist1</listname>
    <subject>This is my list</subject>
    <owner multiple="1">
      <email>first.owner@univ-rennes1.fr</email>
    </owner>
    <owner multiple="1">
      <email>second.owner@univ-rennes1.fr</email>
    </owner>
   </list>
</family>
```

Instantiation with `sympa.pl --instantiate_family test_bundle_of_lists --close_unknown --input_file initial_definition.xml`

And here is /usr/local/sympa/list_data/testlist1/instance.xml:
```/usr/local/sympa/list_data/testlist1/instance.xml
   <list>
    <listname>testlist1</listname>
    <subject>New list subject</subject>
    <owner multiple="1">
      <email>first.owner@univ-rennes1.fr</email>
    </owner>
    <owner multiple="1">
      <email>second.owner@univ-rennes1.fr</email>
    </owner>
    <owner multiple="1">
      <email>third.owner@univ-rennes1.fr</email>
    </owner>
   </list>
```

Issue #693, opened 2019-07-17
Improving data sources


ikedas:
(Part of) Expected changes
----------------------------
* Data sources won't overwrite display name (gecos) of existing subscriber: It is assigned only when a new user is added (included) from data source, or if it is added (subscribed) by users.
* Name of only one data source that was actually included will be shown in subscriber table of web interface.
* Size of `include_sources_subscriber` database field will no longer limit the number of datasources.

See details in the PR #516.

Context
-------
These changes will be included in the next beta (6.2.45b.1).  For more details about changes see the PR page below.

racke:
Why would we limit the size of include_sources_subscriber to varchar(62)? There is no difference in storage for larger field sizes and this field isn't indexed.

ikedas:
> Why would we limit the size of include_sources_subscriber to varchar(62)? There is no difference in storage for larger field sizes and this field isn't indexed.

You says about issue#692.  Currently (until 6.2.44) `include_sources_subscriber` is defined as `varchar(50)` by default.  Indeed the size itself need not limited, but I considered compatibility.

racke:
Compatibility with what?

ikedas:
If the field type was changed to `text`, `sympa.pl --health_check` will complain that database structure is broken.

racke:
varchar(255) wouldn't be text and still a lot larger.

ikedas:
Anyways it isn't big deal with PR#516.

PR #755, opened 2019-09-20
Additional fix to #693

ikedas:
This fixes:
  - a typo mislead by the fix for #693.
  - lack of consideration about column size.

See also [a post on sympa-users](https://listes.renater.fr/sympa/arcsearch_id/sympa-users/2019-09/E830CF905957404CB59FAEF2A5FD302A01ADB335%40CNREXCMBX05P.core-res.rootcore.local).

### DIFF
--- a/src/lib/Sympa/DataSource.pm
+++ b/src/lib/Sympa/DataSource.pm
@@ -94,6 +94,10 @@ sub new {
     my @defkeys = sort keys %{$defopts || {}};
     my @defvals = @{$defopts || {}}{@defkeys} if @defkeys;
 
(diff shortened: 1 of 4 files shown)


racke:
I'm having a hard time to understand this piece of code :thinking:, but thanks for the fix.

racke:
Please raise this to at least 255 characters. Arbitrary limits to varchar doesn't make any sense unless you are indexing the field. Also the web interface should prevent entering strings which are too long to fit into the database.

ikedas:
I chose 50 as length of name of datasource, according to `subject` list parameter.

A short description may be used as short title on web interface, and length above looks acceptable for such usages.

Issue #538, opened 2019-01-26
Support scripts for maintenance


ikedas:
Expected Behavior
-----------------

There may be several scripts to be shared among people interested in maintenance of Sympa source distribution.

Possible Solution
-----------------

Scripts may be put in particular directory in source tree of repository, i.e. `support/`.

Context
-------

Some examples will be shown later.

dverdin:
Hi Soji,
If I understand correctly, correct_msgid just changes the english string, i.e. the translation key, but does not do anything to the translation. Correct?
Is it a way to prepare the files for translation in Pootle?

ikedas:
@dverdin, you are right.  As I wrote on #115:
> If possible, please add your fix to english catalog:
>   - On [English page of translation site](https://translate.sympa.org/en_US/), add correct string as a "translation" of incorrect msgid.
> 
> Your change on the site above will be reflected to source distribution in the future release.

**Note**:
Scripts proposed on this page are expected to be used during maintenance work: None not allowed operation for release work should not use them.  For example with correct_msgid, contributors _must not_ change catalog and sources by it, or catalog and sources might be broken.

 In short, currently it is only me who uses these scripts --- and the person who will take over release work in the future.

dverdin:
OK, that's cool to me. As you're doing the maintenance work, you're probably the best fitted to know what tools are useful.

ikedas:
PR above contains a script:

### pod2md

Converts POD data to Markdown format for documentation site.

This may be used as a replacement of pod2man(1).  To generate Markdown texts of all available PODs, run:
```
$ cd sympa-x.x.x
$ POD2MDOUTPUT=directory 
$ make POD2MAN=/path/to/pod2md
```
then, generated texts will be saved in `directory`.

ikedas:
PR above contains a script:

### xgettext.pl

Extract gettext strings from Sympa source.

This may be used in gettextized Sympa source.  Typically it is invoked as:
```
xgettext.pl --version
xgettext.pl --default-domain=(gettext domain) --directory=$(top_srcdir) \
  --files-from=$(srcdir)/POTFILES.in ...
```

PR #757, opened 2019-09-23
Update gettext support bundled in the source distribution

ikedas:
Update `po/*/Makefile.in.in` to that provided by gettext-tools 0.19.8.

From now on, gettext 0.18.0 or later will be required to update catalog (i.e. msgmerge must have --lang option).

Note: `Makefile.in.in` no longer will be modified, instead edit `Makevars` and/or `xgettext.pl`,


ikedas:
related to #538.

PR #758, opened 2019-09-24
Explicitely setting members visibility to 'noconceal' when the value is undefined in database.

dverdin:
When a user  reviews the list, if she is not allowed to see concealed list members, only the users with a value in the 'visibility_subscriber' database field are actually displayed. Those with an undefined value are not.

The default behaviour in Sympa is that users are visible. Being concealed is an explicit request from the users.

Consequently, I suggest this patch to include a test on undefined values for this field when requesting unconcealed users.

Regards,

David


dverdin:
Darn... My whole merge history appears in the commits. I cancel this PR and redo it correctly.
Sorry.

ikedas:
I roughly suppose fail-safe default would be better to be restrictive, I.e. “conceal”.

dverdin:
Indeed. However, On existing instances, the database contains null value when users have not defined anything AND it means "noconceal" for now.

So either we use the "noconceal" default or we force an explicit value in database during next upgrade (and all the time when modifying a user).

That's way more work.

dverdin:
OK, I did the PR correctly now. Here it is : https://github.com/sympa-community/sympa/pull/760
Let's discuss it there, shall we?

PR #760, opened 2019-09-24
Explicitely setting members visibility to 'noconceal' when the value is undefined in database

dverdin:
When a user reviews the list, if she is not allowed to see concealed list members, only the users with a value in the 'visibility_subscriber' database field are actually displayed. Those with an undefined value are not.

The default behaviour in Sympa is that users are visible. Being concealed is an explicit request from the users.

Consequently, I suggest this patch to include a test on undefined values for this field when requesting unconcealed users.

Regards,

David

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -3297,7 +3297,7 @@ sub get_members {
     if ($role eq 'member') {
         $filter = '';
     } elsif ($role eq 'unconcealed_member') {
-        $filter = " AND visibility_subscriber <> 'conceal'";
+        $filter = " AND (visibility_subscriber <> 'conceal' OR visibility_subscriber IS NULL)";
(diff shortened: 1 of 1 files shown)


ikedas:
[A comment](https://github.com/sympa-community/sympa/pull/758#issuecomment-534485597) by @dverdin :

> Indeed. However, On existing instances, the database contains null value when users have not defined anything AND it means "noconceal" for now.
> 
> So either we use the "noconceal" default or we force an explicit value in database during next upgrade (and all the time when modifying a user).
> 
> That's way more work.

I'm in favor of the latter: At first we force an explicit value in database during next upgrade (and all the time when modifying a user).

dverdin:
Considering my fix was out of what the release manager would favor, I'm closing this PR and try to propose something different in another PR.

PR #775, opened 2019-10-17
Missing `dkim` in send.owner scenario

ldidry:


ikedas:
These also lacks `dkim` authentication method:

- `info.conceal`
- `info.open`
- `info.private`
- `send.ownerauth`

ldidry:
Added, thx!

PR #779, opened 2019-10-25
Make storage into outgoing spool more efficient (#604)

### LINKED ISSUE #604: Unexpected number of sessions regarding nrcpt and avg values
Version
-------
6.2.32

Installation method
-------------------
From @sympa-ja.org repository

Expected behavior
-----------------
We have a list made of 6152 members from 44 different domains. 72 members are from 43 different domains and 6080 are from a single domain. We set nrcpt=1000 and avg=100 globally (not in nrcpt_by_domain.conf)

We would expect Sympa to create 50 sessions with recipients grouped by their destination domain: 7 sessions with the 6080 recipients for the unique domain and 43 sessions with recipients for each one of the other domains.

Actual behavior
---------------
Sympa creates 44 sessions with a number of recipients between 100 and 280 and does not respect the global nrcpt=1000.

Additional information
----------------------
Setting nrcpt in nrcpt_by_domain.conf has no effect.

This issue has for major consequence that a single message turns into the creation of many more messages and reduces deduplication efficiency on the same mail store. Our mail stores receive and store 44 different messages with only 20 to 30 mailboxes referencing each one of these messages, when it should have received and stored only 7 messages with 1000 mailboxes referencing each one of them.

Please tell us how we can help troubleshoot this issue.

Regards,
Frédéric.

ikedas:
Sort recipients to be stored by domains.

This may fix #604.

### DIFF
--- a/src/lib/Sympa/Spool/Outgoing.pm
+++ b/src/lib/Sympa/Spool/Outgoing.pm
@@ -326,6 +326,15 @@ sub _get_recipient_tabs_by_domain {
 
     return unless @rcpt;
 
+    # Sort by domain.
+    @rcpt = map {
+        join '@', grep { defined $_ } @$_;
+    } sort {
+        (($a->[1] // '') cmp ($b->[1] // '')) || ($a->[0] cmp $b->[0])
+    } map {
+        [split /\@/, $_, 2]
+    } @rcpt;
+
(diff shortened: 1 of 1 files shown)


racke:
Sounds indeed like a good idea.

Issue #785, opened 2019-11-07
ldap / ldap 2 level datasource not returning results


xavierba:
Version
-------
6.2.48

Installation method
-------------------
RPM

Expected behavior
-----------------
ldap / ldap 2 level datasource properly returns some mails to include in the list members.

Actual behavior
---------------
No member returned.

Additional information
----------------------
This was working fine with 6.2.44, so I suspect a regression in the datasource rewrite.
Sync'ing the datasource from either the WUI or the CLI doesn't return any error.

```
# sudo -u sympa sympa.pl -d --sync_include some_list@domain.tld
Members of list some_list@domain.tld have been successfully update.
```
Please note the `-d` switch doesn't actually give more verbosity to the output.

Here's the corresponding log:
```
Nov  7 11:30:40 sympa-srv sympa[12016]: info main::_load() Configuration file read, log level set using options: 2
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Conf::cookie_changed() Cookie is stable
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Sympa::Database::new(Sympa::Database, mysql)
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Sympa::Database::connect() Connected to Database Sympa::DatabaseDriver::MySQL <db_name=sympa;db_user=sympa>
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Conf::get_robots_list() Retrieving the list of robots on the server
Nov  7 11:30:40 sympa-srv sympa[12016]: notice main:: Sympa 6.2.48 Started
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Sympa::List::new(some_list@domain.tld, , )
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Sympa::List::load(Sympa::List <>, some_list, domain.tld, ...)
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Sympa::List::sync_include_admin(Sympa::List <some_list@domain.tld>)
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Sympa::List::get_admins(Sympa::List <some_list@domain.tld>, owner,  => )
Nov  7 11:30:40 sympa-srv sympa[12016]: debug2 Sympa::List::sync_include(Sympa::List <some_list@domain.tld>, )
[...]

PR #792, opened 2019-11-08
Patch for a working include_ldap_2level_query (#785)

depouill:
Since 6.2.48, LDAP2 queries don't work anymore.
This is a simple and stupid patch to make it works.


racke:
Possible fix for #785. Thanks for the submission @depouill !

ikedas:
This may fix #785.

xavierba:
I can confirm this fixes #785

bruncsak:
The same problem here, and I can confirm too that this patch fixes it.

PR #830, opened 2019-12-19
Urlize bug fixes

dverdin:
Two bug fixes for Urlize mode:

1- message text was treated as attachment and remained on the server as either "msg.0.bin" or "msg.0.txt" files. Fixed by testing the part type before urlizing it.
2- some special characters in message id led to the attachment to be considered missing by Sympa.

I also added a test file for Message.t. I only test the _urlize_parts function for now but it is still a start!

Cheers,

David


src/lib/Sympa/Message.pm
```diff
@@ -2098,8 +2098,15 @@ sub _urlize_parts {
         return undef;
     }
 
-    ## Clean up Message-ID
-    my $dir1 = Sympa::Tools::Text::escape_chars($message_id);
+    ## Clean up Message-ID and preventing double percent encoding.
```
ikedas: encode_filesystem_safe() would be better to be used instead of ad-hoc escaping.

src/lib/Sympa/Message.pm
```diff
@@ -2145,6 +2152,7 @@ sub _urlize_one_part {
         $filename = Encode::encode_utf8($filename)
             if Encode::is_utf8($filename);
     } else {
+        return undef if ($entity->effective_type =~ m{\A(multipart|text)});
```
ikedas: Attachment with text/plain type should be urlized. Nested multipart should also urlized.

racke:
@dverdin can you please get rid of all the unrelated commits in the PR?

racke:
"Fixing merge conflicts" is certainly something which shouldn't be in a PR.

dverdin:
Hrm... Certainly... I must admit I don't have any idea how to do this. Any pointer you could provide for this?
Thanks!

dverdin:
OK. A lot of these commits are useless. I'm trying to remove them using "rebase".
I'll keep you updated.

dverdin:
OK. Now it's clean. Sorry for the mess.

src/lib/Sympa/Message.pm
dverdin: Indeed. I never noticed this function.
Fixed in commit #c8195a6

src/lib/Sympa/Message.pm
dverdin: I'm not sure actually. But I might have a biased opinion.
On my opinion, the main point of urlized is to decrease the size of mails sent. Consequently, I want to get rid of images, PDF and all theses attachments that are heavy.
Text is rarely that heavy. So if we keep all mulitpart and text parts, but get rid of the rest, we have really lighter messages which is good enough for me, as we would keep all the text that gives sens to the message.
However, I might have misunderstood the way _urlize_one_part works and it may ignore the content of nested multipart. In that case, this commit needs reworking. I'll add some tests about this situation.

dverdin: OK. Inded, if I ignore a multipart, it won't go recursively inside it to extract attachments.
So I think I need to treat "multipart/mixed" differently.

dverdin: I refactored the code slightly to allow recursive urlization of nested multiparts.

dverdin:
There are still problems to be fixed with nested multiparts.
More on this later.

dverdin:
Now it looks good.

- I'm handling nested multiparts,
- I'm using encode_filesystem_safe

The code is running on our production instance. I'll now soon if some cases are still missing.

dverdin:
Considering I had to amend a lot of things to my original PR, I'll close it and propose it again in another PR, so it is easily readable.
And I'll add the test to travis.

PR #840, opened 2020-01-10
Urlize mode bug fixes

dverdin:
Fixing bugs in urlize mode : 

1- message text was treated as attachment and remained on the server as either "msg.0.bin" or "msg.0.txt" files. Fixed by testing the part type before urlizing it.
2- some special characters in message id led to the attachment to be considered missing by Sympa.

I also introduce a test file for Message.t, testing only urlization. this test is set in travis and the test succeeds. Note that commits 065312d and fcadbac are related to travis. The first one to enable the test, the second one to have Sympa::Constants without running make.

### DIFF
--- a/src/lib/Sympa/Message.pm
+++ b/src/lib/Sympa/Message.pm
@@ -2088,7 +2088,7 @@ sub _urlize_parts {
 
     ## Only multipart/mixed messages are modified.
     my $eff_type = $entity->effective_type || 'text/plain';
-    unless ($eff_type eq 'multipart/mixed' or $eff_type eq 'multipart/alternative') {
+    unless ($eff_type eq 'multipart/mixed' or $eff_type eq 'multipart/alternative' or $eff_type eq 'multipart/related') {
         return undef;
     }
 
@@ -2106,7 +2106,7 @@ sub _urlize_parts {
             $expl, $dir1);
         return 0;
     }
-    return _urlize_sub_parts($entity, $list, $message_id, $dir1, 0, 0);
+    return _urlize_sub_parts($entity, $list, $message_id, $dir1, 0);
 }
 
 sub _urlize_sub_parts {
@@ -2114,22 +2114,22 @@ sub _urlize_sub_parts {
     my $list       = shift;
     my $message_id = shift;
     my $directory = shift;
-    my $is_sub_part = shift;
     my $i = shift;
     my @parts = ();
     use Data::Dumper;
+    my $parent_eff_type = $entity->effective_type();
     foreach my $part ($entity->parts) {
         my $eff_type = $part->effective_type || 'text/plain';
         if ($eff_type eq 'multipart/mixed') {
-            my $p = _urlize_sub_parts($part->dup, $list, $message_id, $directory, 0, $i);
-            push @parts, $p;
             $i++;
-        } elsif ($eff_type eq 'multipart/alternative' and !$is_sub_part) {
-            my $p = _urlize_sub_parts($part->dup, $list, $message_id, $directory, 1, $i);
+            my $p = _urlize_sub_parts($part->dup, $list, $message_id, $directory, $i);
             push @parts, $p;
+        } elsif (($eff_type eq 'multipart/alternative' or $eff_type eq 'multipart/related') and $i < 2) {
             $i++;
+            my $p = _urlize_sub_parts($part->dup, $list, $message_id, $directory, $i);
+            push @parts, $p;
         } else {
-            my $p = _urlize_one_part($part->dup, $list, $directory, $i);
+            my $p = _urlize_one_part($part->dup, $list, $directory, $i, $parent_eff_type);
             if (defined $p) {
                 push @parts, $p;
                 $i++;
@@ -2144,10 +2144,13 @@ sub _urlize_sub_parts {
 }
 
 sub _urlize_one_part {
-    my $entity = shift;
-    my $list   = shift;
-    my $dir    = shift;
-    my $i      = shift;
+    my $entity      = shift;
+    my $list        = shift;
+    my $dir         = shift;
+    my $i           = shift;
+    my $parent_eff_type    = shift;
+    
+    return undef unless ($parent_eff_type eq 'multipart/mixed');
 
     my $expl     = $list->{'dir'} . '/urlized';
     my $listname = $list->{'name'};
--- a/src/lib/Sympa/Message.pm
+++ b/src/lib/Sympa/Message.pm
@@ -2088,7 +2088,7 @@ sub _urlize_parts {
 
     ## Only multipart/mixed messages are modified.
     my $eff_type = $entity->effective_type || 'text/plain';
-    unless ($eff_type eq 'multipart/mixed') {
+    unless ($eff_type eq 'multipart/mixed' or $eff_type eq 'multipart/alternative') {
         return undef;
     }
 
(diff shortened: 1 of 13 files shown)


dverdin:
All remarks made in PR #830 are included in the current PR.

t/stub/Sympa/Constants.pm
```diff
@@ -6,7 +6,35 @@
 
 package Sympa::Constants;
 
-use constant LOCALEDIR => 't/locale';
+use constant VERSION => '6.2.52';
+use constant USER    => $ENV{LOGNAME} || $ENV{USER} || getpwuid($<);
+use constant GROUP   => $ENV{LOGNAME} || $ENV{USER} || getgrgid($<);
```
ikedas: - `getpwuid` and `getgrgid` are executed in list context and don't give single name of user/group.
- `$ENV{LOGNAME}` is not an actual user name (e.g. if superuser changed their own user by `su`, login name won't change).
- `$ENV{LOGNAME}` and `$ENV{USER}` are not appropriate for user's group.

Thus IMHO:
``` perl
use constant USER    => $ENV{USER} || [getpwuid($<)]->[0];
use constant GROUP => [getgrgid([getpwnam(USER)]->[3])]->[0];
```

dverdin: Thanks @ikedas !

I tested it and it works as well as before, so I see no reason not to comply to your suggestion.

So I followed your recommendation in commit d911ef8.

dverdin: A remark: on another branch, the build completes only when I use the following:

`use constant USER    => [getpwuid($<)]->[0];`

Just as if `$ENV{USER}` returned an unusable value. Do you think it is necessary?

ikedas: I don't know whether `$ENV{USER}` is useful. It is often set by the shell, but not always.

dverdin: I feel that [getpwuid($<)]->[0] is enough in the case of testing.

ikedas: If it was,
``` perl
use constant USER    => [getpwuid $<]->[0];
use constant GROUP   => [getgrgid $(]->[0];
```
may be enough.

dverdin: OK. I'll fix this. Any other remarks?

ikedas:
I think simply recursive urlization of subparts is not always appropriate.

We would be better to consider more precise cases where the subparts are treated as "attachment":

  - multipart/mixed

    This type on the top level may be special. Subparts may be treated as either ---
      - The first text/* part should be treated as "text", and the others should be "attachment" even if it is a text.
      - Or, text/* part that is decodable (i.e. with proper charset parameter) and does _not_ have "content-disposition: attachment" should be treated as "text", the others should be "attachment".

  - multipart/alternative, multipart/related
    Urlizing any subparts is not useful, because MUA is expected to show only most suitable subpart, or to render message using all subparts.

  - multipart/digest, multipart/report etc.
    Urlizing any subparts is not useful, because a multipart itself is an integrated structure.

  - message/rfc822, message/global
    Urlizing any parts this part contains is not useful, because this type represents embedded e-mail message and its content should not be broken.

Thus, IMO only multipart/mixed may be modified recursively (with restrictions described in above) to "urlize" messages.

dverdin:
Great review @ikedas !
I'll confront it to the problems I met in the sample files provided in the PR to check whether your suggestions work with them. These were originally messages for which urlization did not work correctly.
More news in a day or two!
David

t/Message.t
```diff
@@ -52,6 +52,12 @@ make_path($home_dir);
 dircopy('t/data/list_data/', $home_dir);
 make_path($etc_dir);
 
+## Redirecting standard error to tmp file to prevent having logs all over the output.
+open my $fh, '>', "$tmp_dir/error_log" or die "Can't open file $tmp_dir/error_log in write mode";
+close(STDERR);
+my $out;
+open(STDERR, ">>", \$out) or do { print $fh, "failed to open STDERR ($!)\n"; die };
+
```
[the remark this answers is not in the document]
dverdin: Really? Because I found loads of logs in the Travis output.
Anyway, it is annoying when running the single test with prove.
Is it is not bug-prone, I suggest to keep it, to make testing smoother.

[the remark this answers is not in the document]
ikedas: I think the outputs to stderr are useful. I don't agree to throw them away into trashbox file.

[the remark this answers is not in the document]
ikedas: To control log output:
``` perl
my $log = Sympa::Log->instance;
$log->{log_to_stderr} = 'err';
```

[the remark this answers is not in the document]
dverdin: Cool ! Thanks !

dverdin:
Okie dokie... I had a look and I still have a problem with nested multiparts.

Apple Mail sometimes sends emails with attachments within multipart/alternative messages.
In those messages, the first part is text/plain, the second is multipart/mixed, and contains the text/html version of the first part and any attachments.

I think this is wrong because it makes no sense to put the attachments within an alternative part, however it is a fact. I'd like to simply ignore such misbehaviours but end users will probably consider it is Sympa's fault is urlize mode does not work.

I'd the like to consider multipart/related and multipart/alternative as potential attachments bearers.

That would lead to the following (changes to your version are in bold font):

>   - multipart/mixed
> 
>     This type on the top level may be special. Subparts may be treated as either ---
>       - ~~The first text/* part should be treated as "text", and the others should be "attachment" even if it is a text.~~ **(I prefer the scond option, simpler to implement)**
>       - text part that is decodable (i.e. with proper charset parameter) and does _not_ have "content-disposition: attachment" should be treated as "text", the others should be "attachment".
> 
>     --- and only "attachment" can be urlized.
> 
>   - multipart/alternative, multipart/related
> **Don't urlize any subpart except if it is a multipart/mixed.**
>     Urlizing any subparts is not useful, because MUA is expected to show only most suitable subpart, or to render message using all subparts.   Entire part (multipart) should be urlized if necessary.
> 
>   - multipart/digest, multipart/report etc.
>     Urlizing any subparts is not useful, because a multipart itself is an integrated structure.   Entire part (multipart) should be urlized if necessary.
> 
>   - message/rfc822, message/global
>     Urlizing any parts this part contains is not useful, because this type represents embedded e-mail message and its content should not be broken.   Entire part (message) should be urlized if necessary.
>

dverdin:
OK, I have implemented what we discussed. A slight change : I urlize sub-parts of multipart/alternative, whichever they are, bu only one level below the message top level part. That means that if they contains anything other than valid text, it will be urlized. That leaves a chance to catch a multipart/mixed inside, but we won't go further.

This is not satisfactory yet, though it works well with the test files. It still needs working.

I have reorganized the test files to make their usage more explicit.

dverdin:
OK, now it does exactly what I wanted. Only going to urlize sub parts of a multipart/related if it is a multipart/mixed.
I'll wait to see whether all the tests complete, then maybe squash some commits to make the whole PR more understandable.

dverdin:
OK, I squashed the commits to keep only four commits, combining the code modifications we compromised over. It should be easier to understand fr future developers.

src/lib/Sympa/Message.pm
```diff
@@ -2145,6 +2162,15 @@ sub _urlize_one_part {
         $filename = Encode::encode_utf8($filename)
             if Encode::is_utf8($filename);
     } else {
+        my $content_disposition = $entity->head->get('Content-Disposition');
```
ikedas: Use ``mime_attr()`` to handle things like "``Content-Disposition: (comment) Attachment (comment)``":
```perl
         my $content_disposition = lc($entity->head->mime_attr('Content-Disposition') // '');
```
and no regexp is required to compare it with ``'attachment'``.

dverdin: Sure? No spaces after/before the field content?
In that case, I'll do as you suggest.
Regards,

ikedas: get() gets value of the field, ie. “ `(comment) Attachment (comment); filename="name.ext"`”. mime_attr() gets specified attribute in the field, ie. “ `Attachment `”

dverdin: Thanks @ikedas !
I really need to dig deeper into MIME::Tools.
Fixed in commit 72451a0.
Cheers,

dverdin:
Actually, I still found some cases in production, just today, that don't come out of urlization as expected.
So I'll have to dig in them a little to understand why my code don't process them correctly.
I'll post an update here soon.

Regards,

dverdin:
OK, now that seems right. I had to go one level further in the message architecture but I added an explicit rule in _urlize_one_part that will prevent urlization to occur when the parent part is not a multipart/mixed.
As usual, I put this code in production right now.

ikedas:
I merged. Thanks @dverdin !

dverdin:
Thanks @ikedas !
Glad to see this behind me...
Now, I'm heading towards the next bunch of PR...
Regards,
David

Issue #872, opened 2020-02-17
Shared document configuration in edit list even if feature turned off


racke:
Version
-------
6.2.52

Expected behavior
-----------------

Configuration in "Privileges" shouldn't be visible and possible when _shared_feature_ is `off`.

Additional information
----------------------
![test   Test   edit_list_request](https://user-images.githubusercontent.com/143678/74639442-370be400-516e-11ea-9731-ad114210fc1c.png)

khaoula-abaidi:
The shared_feature off into sympa.conf is not accepted with 6.2.40 version.
How can i disable the shared repositories options for all users?

ikedas:
@khaoula-abaidi , please submit the new issue for ypur question.

khaoula-abaidi:
> The shared_feature off into sympa.conf is not accepted with 6.2.40 version.
> How can i disable the shared repositories options for all users?

@ikedas how can i submit it ?

ikedas:
@khaoula-abaidi ,

> @ikedas how can i submit it ?

![image](https://user-images.githubusercontent.com/5743546/99503774-f58cbb80-29c1-11eb-91bc-a3b2c946e193.png)

Please click "New issue" button above in this page.

PR #877, opened 2020-02-18
Skip edit for shared document options when disabled (#872)

racke:
Fixed and tested, albeit kind of ugly.


ikedas:
PR 899 looks promising. Close this PR without merge.

PR #884, opened 2020-02-20
Update bug tracker url in serveradmin section.

racke:


ikedas:
I suggest [CONTRIBUTING](https://github.com/sympa-community/sympa/blob/sympa-6.2/CONTRIBUTING.md) text (Should it be added to Jekyl site?).

PR #891, opened 2020-02-26
URLize: Use filesystem-independent escaping for names of files stored

ikedas:
An insignificant bug.

Discouraged `Sympa::Tools::Text::escape_chars()` was used in the name of file stored for URLize mode.


ikedas:
@dverdin , thanks for your review!

PR #898, opened 2020-03-02
Skip edit for shared document options when disabled (#872) [2]

ikedas:
Alternative to #877 .


racke:
Yes, that's good. And I think we can use this for other purposes as well. :+1:

ikedas:
PR 899 looks promising. Close this PR without merge.

PR #900, opened 2020-03-04
Some bugs related to urlize mode

ikedas:
* URLize: If personalization was enabled, urlization is prevented.  Because directory for attachment files will have been created by test_personalize() and can't be created again.
* Suppress unuseful logs skipping urlization of attachments with Content-Disposition.
* WWSympa: do_attach(): Avoid urlized attachments to be rendered by the browser.


dverdin:
Excellent. I did not notice this problem with merge.
Thanks Soji!

PR #903, opened 2020-03-04
List admin update through the command line

dverdin:
See Issue #524 for discussion.

The aim of this PR is to give administrators the possibility to update list administrators (owners or editors) through the command line. The idea here is to be able to change details of a list administrator (visibility, privileges, etc.) in one operation, witout having to delete, then add it. It alos allows to replace a user by another one, keeping all previous details except if told otherwise.
As it is based on a Request handler, it also introduces the possibility to do so through the Sympa API (SOAP, and later REST).
It is complemented by two other PR, allowing deletion and addition of list admins.

Adding :
    - a Request handler dedicated to list admin update,
    - a sympa.pl command line option to update a list admin, based on this handler,
    - tests for the handler module.

The command line allows the following options:

```--update_list_admin --current_email=user@example.com (--list=list@domain | --robot=robot ) [ --new_email=new_user@example.com --role=role --profile=privileged|normal --info=<description> --visibility=conceal|noconceal --reception=mail|nomail --gecos=<a gecos> ]```

Updates a list admin, either replacing it with a new user (if "--new_email" is given and is different from "--current_email") or simply updating her informations.

    - "--current_email" is the list admin's email address. It is mandatory.
    - Either "--list" or "--robot" is mandatory.
    - If no "--role" is given, all administrative roles will be updated (both editor and owner).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,6 +9,7 @@
     - coverage-setup
     - cpanm --quiet --notest --installdeps --with-develop --with-feature=Data::Password --with-feature=ldap --with-feature=safe-unicode --with-feature=smime --with-feature=soap --with-feature=sqlite .
     - cpanm --notest --quiet Unicode::CaseFold
+    - cpanm --notest --quiet DBD::SQLite
     - autoreconf -i
     - ./configure
     - cd src; make; cd ..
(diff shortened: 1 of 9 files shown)


ikedas:
Currently updating parameters by each are restricted/allowed according to settings in `edit_list.conf`.  `update_list_admin` handler doesn't reflect these settings.

dverdin:
Indeed!
That should be fixed.
Which put the conversation about rethinking the handler on the table, doesn't it?

ikedas:
Superseded by #1503.

PR #906, opened 2020-03-06
Fix logging of open file error in Config_XML's constructor.

racke:


racke:
I prefer the first approach, as it improves clarity for casual code lurkers :-). Code adjusted. Thanks for the help @ikedas

Issue #90, opened 2017-10-11
Reorganizing log levels


ikedas:
Currently, `err` and `notice`, or `info` and `notice` levels are sometimes confused. And, `err` level outputs traceback to syslog even when the error was caused by user.

So I wish to clarify the purpose of each log level.

See also issue #64.

Summary of proposal
-------------------

  * `err` level will be separated to `crit`, `intern` and `user`.

  * `user` level will be used for information caused by users' action.
    `crit`, `intern` and `notice` levels will be used for the system.

  * `notice` and `info` level will be used for successful result.

  * `debug*` levels will be used for things not corresponding to any levels above.

  * `trace` level will be deprecated.

Comparison
-----------

Logging levels are *approximatily* associated as the table below.

| Syslog (*1) | Report (*2) | This proposal | comment |
|----------|----------|----------|----------------------------------|
| `err`    |  -       | `crit`   | When process died                |
| `err`    | `intern` | `intern` |                                  |
| `err`    | `auth`   | `user`   | Authorization by scenario failed |
| `err`    | `user`   | `user`   |                                  |
| `notice` |  notice  | `notice` |                                  |
| `info`   |  -       | `info`   |                                  |
| `debug*` |  -       | `debug*` |                                  |
| `trace` | -        | -          |                                  |

(*1) Sympa::Log::syslog() and wwsympa's wwslog().
(*2) Sympa::Report::reject_report_web() and Sympa::Report::notice_report_web().

Detailed description
--------------------

### `crit`: Non-recoverable error: Program cannot continue processing.

In other words, Sympa has no choice but to die.

e.g.:
- Mandatory file or directory (such as spool) does not exist, cannot be open or not be writable.
- Unrecoverable errors in `sympa.conf` or `robot.conf` (such as malformed parameter value without default).
- Program died (including the case that `die()` was explicitly executed).
- and other failure to prevent running the program.

n.b. This level outputs traceback to log and STDERR.

### `intern`: Internal server error

Precisely, recoverable failures corresponding to foreign component (database, MTA, DNS, ...).

e.g.:
- Submitting one or more messages to MTA failed.
- Logging failed.
- Database server was (temporarily or permanently) gone.
- Query was denied by SQL or LDAP server.
- DNS returned malformed response (DKIM etc.).
- CAS server unexpectedly denied ticket.
- Synchronization of list members unexpectedly failed.
- ... and other errors interacting with foreign components.

n.b. This level outputs traceback to log.

[...]

PR #908, opened 2020-03-07
Improve logging of archive errors due to possible permission problems 

racke:


src/lib/Sympa/Archive.pm
```diff
@@ -316,8 +316,9 @@ sub html_next {
     unless ($self->{_html_metadatas}) {
         my $dh;
         unless (opendir $dh, $self->{arc_directory}) {
-            die sprintf 'Cannot open dir %s: %s', $self->{arc_directory},
-                $ERRNO;
+            $log->syslog('err', 'Cannot open dir %s: %s', $self->{arc_directory},
+                         $ERRNO);
+            return undef;
```
ikedas: I think this is the case when "Sympa has no choice but to die" (cf. #90 ).  It's fatal that `$self->{arc_directory}` is set, i.e. `select_archive()` has succeeded, while it cannot be opened.

src/lib/Sympa/Archive.pm
```diff
@@ -215,9 +215,16 @@ sub html_fetch {
     return undef unless $self->{arc_directory};
     return undef unless $options{file};
 
-    my $handle =
-        IO::File->new($self->{arc_directory} . '/' . $options{file}, '<');
-    return undef unless $handle;
+    my $html_file = $self->{arc_directory} . '/' . $options{file};
+    my $handle = IO::File->new($html_file, '<');
```
ikedas: I don't know why older men used `IO::File` instead of plain file handle. Though there seems no problem...

racke: Yes, it would make sense to always use the same pattern. But that's outside the scope of this pull request.

src/lib/Sympa/Archive.pm
racke: I can see your point ... but it leaves the web interface user with the die output which should not happen other than real corner cases. What about checking whether the directory can be read in the beginning of do_arc?

ikedas: Checking readability may not be the solution, I think. But this topic seems beyond the scope of this pr, and I’ll stop commenting for the present.

PR #911, opened 2020-03-14
Add unsubscribe command to remove emails from list in sympa.pl

kukoarmas:
In the sympa.pl CLI we have the "import" option to add subscribers to a list, but we lack a similar command to remove subscribers from the list.
This command is useful if you need to remove subscribers from a batch script

### DIFF
--- a/src/sbin/sympa.pl.in
+++ b/src/sbin/sympa.pl.in
@@ -66,6 +66,7 @@ unless (
         'config|f=s',                   'lang|l=s',
         'mail|m',                       'help|h',
         'version|v',                    'import=s',
+        'unsubscribe=s',
         'make_alias_file',              'lowercase',
         'sync_list_db',                 'md5_encode_password',
         'close_list=s',                 'rename_list=s',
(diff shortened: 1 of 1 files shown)


ikedas:
For me, this PR seems related to #901/#902. IMHO `--add` and `--del` (or maybe `--add_member` and `--del_member`) would be better to be implemented.

kukoarmas:
> For me, this PR seems related to #901/#902. IMHO `--add` and `--del` (or maybe `--add_member` and `--del_member`) would be better to be implemented.

Do you mean that I should implement those two commands? If I implement --add_member for multiple entries, wouldn't it duplicate the --import functionality?

I'm ok implementing both commands, I'm just asking ;)

ikedas:
> Do you mean that I should implement those two commands? If I implement --add_member for multiple entries, wouldn't it duplicate the --import functionality?

No you may not, if you will wait for the work by @dverdin progressing.

N.B. I think the “--import“ option On CLI was introduced for consistency with “Multiple add” function on web GUI.  The “del” function looks not having multiple counterpart.

ikedas:
@kukoarmas , sorry for delayed merge.  As @dverdin 's work got stuck, this PR was merged first.

The name of command was changed to "del" standing for the action by admins to delete any subscribers.

Thank you for your contribution!

PR #912, opened 2020-03-15
Data sources: ldap_2level: Returns at most one result not according to select2 parameter

### LINKED ISSUE #893: ldap_2level_query "select2 all" is returning only one result
In a dynamic data source, a ldap_2level_query seems to return only one result when "select2 all" is used. 

I have a list with a "include_ldap_2level_query" statement in the config file. I am only getting partial results on the second select. 

Example : I have 20 students, I want to get the parents for every student. I should have 40 parents but the result only returns 20 parents. It seems there is no difference between "select2 first" and "select2 all" in 6.2.52

I am on RHEL7 and the LDAP source is Microsoft Active Directory.

The bug is the same if I connect with LDAP with STARTTLS.

I have 6.0.1 in a lab and its working fine, I have the full results with the same config fille.

Version
-------
6.2.52

Installation method
-------------------
rpm

Expected behavior
-----------------
"select2 all" should return all results from a filter

Actual behavior
---------------
"select2 all" return only the first result

Additional information
----------------------
include_ldap_2level_query
select2 all
scope2 sub
suffix2 OU=PARENTS,OU=PEOPLE,dc=[redacted]
suffix1 OU=STUDENTS,OU=PEOPLE,dc=[redacted]
attrs1 cn
select1 all
filter2 (kids=[attrs1])
timeout1 60
name parent023706
attrs2 mail
bind_dn CN=[redacted]
scope1 sub
port 389
host [redacted]
bind_password [redacted]
filter1 (&(departmentNumber=023)(businessCategory=706))
timeout2 60

ikedas:
This may fix #893 .


ikedas:
I use [Test::Net::LDAP](https://metacpan.org/pod/Test::Net::LDAP::Util), added it to "on 'develop'" block in `cpanfile` and placed its wrapper`Sympa::Test::MockLDAP` under `src/lib` directory. They are used only when tests are performed and not used in run time.

PR #914, opened 2020-03-15
`sympa.pl --sync_include`: No errors were reported to caller

### LINKED ISSUE #907: Sync include from commandline oblivious of errors
Errors happening during synchronization of data sources don't bubble up to the caller.
```

/usr/local/sympa/bin/sympa.pl --sync_include=bart@lists.example.com
Members of list bart@lists.example.com have been successfully updated.

```
It doesn't matter whether there is an invalid `.incl` file or a SQL problem, the hapless user thinks everything is order.

ikedas:
`sympa.pl --sync_include`: No errors weren't reported to caller, even if one or more of data sources failed inclusion.

This PR depends on PR #905 .

This PR may fix #907 .

### DIFF
--- a/default/mail_tt2/report.tt2
+++ b/default/mail_tt2/report.tt2
@@ -384,6 +384,9 @@
         [%|loc(report_param.name,report_param.listname)%]Including subscribers from data source "%1" into list %2 failed.[%END%]
     [%~ END %]
 
+[%~ ELSIF report_entry == 'include_incomplete' ~%]
+   [%|loc(report_param.result.added,report_param.result.held,report_param.result.updated)%]Inclusion partially or entirely failed (%1 added, %2 held, %3 updated).[%END%]
+
 [%~ ELSIF report_entry == 'include_performed' ~%]
    [%|loc(report_param.result.added,report_param.result.deleted,report_param.result.updated)%]Inclusion succeeded (%1 added, %2 deleted, %3 updated).[%END%]
 
--- a/default/mail_tt2/report.tt2
+++ b/default/mail_tt2/report.tt2
@@ -357,7 +357,37 @@
 [%~ ELSIF report_entry == 'no_message_to_moderate' ~%]
 [%|loc(report_param.listname)%]There is no message to moderate for list '%1'.[%END%]
 
-[%~ ELSIF report_entry == 'time_limit_for_include' ~%]
+[%~ ELSIF report_entry == 'include_skip' ~%]
+    [% IF report_param.role == 'owner' ~%]
+        [%|loc(report_param.listname)%]Including list owners into list %2 was skipped.[%END%]
+    [%~ ELSIF report_param.role == 'editor' ~%]
+        [%|loc(report_param.listname)%]Including moderators into list %2 was skipped.[%END%]
+    [%~ ELSE ~%]
+        [%|loc(report_param.listname)%]Including subscribers into list %2 was skipped.[%END%]
+    [%~ END %]
+
+[%~ ELSIF report_entry == 'include' ~%]
+    [% IF report_param.role == 'owner' ~%]
+        [%|loc(report_param.name,report_param.listname)%]List owners were included from data source "%1" into list %2.[%END%]
+    [%~ ELSIF report_param.role == 'editor' ~%]
+        [%|loc(report_param.name,report_param.listname)%]Moderators were included from data source "%1" into list %2.[%END%]
+    [%~ ELSE ~%]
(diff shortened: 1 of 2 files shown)


src/lib/Sympa/Request/Handler/include.pm
```diff
@@ -280,7 +292,7 @@ sub _twist {
         $request, @result{qw(added deleted updated)}
     );
     $self->add_stash($request, 'notice', 'include_performed',
```
racke: This notice will be shown regardless of the outcome of the inclusion:

```
sympa@local-sympa:/var/lib/sympa/list_data/lists.example.com/test$ /usr/local/sympa/bin/sympa.pl --sync_include=test@lists.example.com
include [notice] Including subscribers from data source "subscribers.txt" into list test failed.
include [notice] Inclusion succeeded (0 added, 0 deleted, 0 updated).
```

racke:
It would be also useful to add "skipped" to the result hash so the user is aware of bad data from the inclusion source.

src/lib/Sympa/Request/Handler/include.pm
ikedas: Yes.  If multiple data sources are configured for a single list, some may succeed and some may fail.  `include_performed` report simply informs that inclusion task has been finised. What is more suitable text? "Inclusion finished"?

ikedas:
> It would be also useful to add "skipped" to the result hash so the user is aware of bad data from the inclusion source.

`added`, `deleted` and `updated` are total numbers of records (users) successfully processed. I think number of `skipped` records can't be known.

ikedas:
> It would be also useful to add "skipped" to the result hash so the user is aware of bad data from the inclusion source.

@racke, though I don't know this is what you suggested, I added numb er of "held" users: See cd429cb.

src/lib/Sympa/Request/Handler/include.pm
ikedas: I separated `include_incomplete` and `include_performed`. See cd429cb.

default/mail_tt2/report.tt2
```diff
@@ -384,6 +384,9 @@
         [%|loc(report_param.name,report_param.listname)%]Including subscribers from data source "%1" into list %2 failed.[%END%]
     [%~ END %]
 
+[%~ ELSIF report_entry == 'include_incomplete' ~%]
+   [%|loc(report_param.result.added,report_param.result.held,report_param.result.updated)%]Inclusion partially or entirely failed (%1 added, %2 held, %3 updated).[%END%]
+
```
racke: I would prefer to break this in two cases (partially / entirely).

We can determine whether a inclusion was partially or entirely failed, so I don't see a point in mixing them together. I also assume that in most cases we have a single datasource, so no need to confuse these users with partially failures.

ikedas: Partial and entire cases were separated. See ee42f48

racke:
There seems to be a logic error. So I have an inclusion from a file which doesn't exist:

```
Mar 20 08:42:11 local-sympa sympa[11984]: notice Sympa::Spindle::ProcessRequest::_twist() Processing Sympa::Request <action=include;context=test@lists.example.com;role=member>
Mar 20 08:42:11 local-sympa sympa[11984]: err main::#846 > Sympa::Spindle::spin#95 > Sympa::Request::Handler::include::_twist#185 > Sympa::Request::Handler::include::_update_users#343 > Sympa::DataSource::open#134 > Sympa::DataSource::File::_open#40 Unable to open file "/var/lib/sympa/list_data/lists.example.com/test/subscribers.txt": No such file or directory
Mar 20 08:42:11 local-sympa sympa[11984]: info Sympa::Request::Handler::include::_twist() Sympa::Request <action=include;context=test@lists.example.com;role=member>: Success, 0 added, 0 deleted, 0 updated
```

The error happens in "# II. Include new entries." but you do the accounting in " # III. Expire outdated entries.". 

I also think failed data sources should be excluded from further steps, thus remove them from them from _@dss_ before going to the expiration step.

ikedas:
> The error happens in "# II. Include new entries." but you do the accounting in " # III. Expire outdated entries.".

Step III (expiration) can happen only once in one execution of include request, even if there are multiple data sources.  Because, if a user will no longer be included by updating of one data source, they can be included by another data source

ikedas:
I re-pushed. @racke, could you please check the new patches? Thanks.

racke:
Complete failure and success are looking good to me now. I will try to test partial failure as well.

ikedas:
@racke, did you find anything signing problem? If you don't have complaint, could you please merge this pr?

PR #916, opened 2020-03-18
Replace `http_host` parameter with `wwsympa_url_local` parameter (addition to #910)

ikedas:
This PR is based on #910.

Though the `http_host` parameter has been obsoleted, many site may set this parameter as the same as host part of `wwsympa_url` and they may have to remove it (e.g. see [a post on dev list](https://listes.renater.fr/sympa/arcsearch_id/sympa-developpers/2020-03/cef874de-1935-36d2-fa46-0b6ac8c309d8%40linuxia.de)).

Instead of keeping this complicating parameter, we'd be better to introduce the new `wwsympa_url_local` parameter to specify proxy upstream URL.

### DIFF
--- a/src/lib/Sympa/ConfDef.pm
+++ b/src/lib/Sympa/ConfDef.pm
@@ -1377,10 +1377,9 @@ our @params = (
         'gettext_comment' =>
             'This is used to construct URLs of web interface.',
     },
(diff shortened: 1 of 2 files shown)


racke:
Thanks for the quick fix @ikedas

PR #923, opened 2020-03-29
Fix "Latests" typo

racke:


ikedas:
[Fixed](https://translate.sympa.org/en_US/sympa/translate/en_US.po#unit=59925).

Issue #466, opened 2018-11-07
Commas in source_parameters


junior013:
Expected Behavior
-----------------
Write commas in any parameter in sorce_parameters list (member_include, owner_include, moderator_include)

Current Behavior
----------------
source_parameters string simply splitted at commas, even them inside quotation or escaped
`my @data = split(',', $parsing->{'data'});` (

Possible Solution
-----------------
Change source_parameters format to JSON or use multiline textbox, one parameter per line, split this string at newlines instead commas

Context
-------
I try to use LDAP (AD) templates where filter is a parameter, but filter string like "memberOf=CN=somegroup,OU=example,OU=com" can't work :/

xandm:
May I suggest the following solution, which is a small change to the code and allows commas to simply be escaped with a backslash:

```
use Text::Parsewords;
...
my @data     = quotewords(',', 0, $entry->{'source_parameters'})
```

ikedas:
I propose a new parameter (say, ``source_parameter``) with multiple occurrences, for example:
``` code
include_ldap_query
  source_parameter memberOf=CN=somegroup,OU=example,OU=com
  source_parameter something other
  ...
```

ikedas:
Contrary to expectations, it turned out that my proposal above needs a bit large amouont of changes on code.  I'm sorry I will popstpone related PR to the next of next stable (maybe 6.2.62 in Spring).

dverdin:
That will be a good solution indeed.
In the meantime, applying @xandm patch solves the problem: you can escape commas with a backslash.

ikedas:
I'm working for the fix [here](https://github.com/sympa-community/sympa/compare/6.2.60...ikedas:issue-466).  However I ask for postponement again, because it turned out that it needs big change on UI code; it'll take some more time.

salaun-urennes1:
Any update on this issue?

We have heavy use of `member_include` on our Sympa server. This limitation, regarding use of comma in 		`source_parameters`, is really annoying when it comes to SQL queries.

We have this kind of query : `select mail from table1, table2 where table1.id=table2.id ...`
We end up with SQL query truncated after comma.

Alternatives are:
1. setting `include_sql_query` parameter in list config, but we have around 50 mailing lists querying that database and we wish to keep the database parameters (and password) in a single place;
2. defining SQL views (to get rid of SQL joins), but it is a tedious job.

We ended up patching Sympa with the code suggested by Xand; it solved the problem.
Than you Xand :-)

[...]

PR #940, opened 2020-05-14
Allow escaped commas in source_parameters

xandm:
Relates to https://github.com/sympa-community/sympa/issues/466

This allows LDAP DNs etc to be used in `source_parameters`, for example:

`source_parameters CN=somegroup\,OU=example\,OU=com`

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -34,6 +34,7 @@ use English qw(-no_match_vars);
 use IO::Scalar;
 use POSIX qw();
 use Storable qw();
+use Text::ParseWords;
 
 use Sympa;
 use Conf;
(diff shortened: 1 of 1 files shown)


ikedas:
At least reading [description about Text::ParseWords](https://perldoc.perl.org/Text/ParseWords.html), I guessed that your change will also allow the format such as:
``` code
source_parameters "CN=somegroup,OU=example,OU=com"
```

ikedas:
I feel the behavior above is nice.  However, if we do the things like:
```
source_parameters CN=somegroup"

source_parameters "CN=somegroup
```
input will be silently ignored. This behavior looks slightly astonishing for users.  How can we report such cases as errors?

I checked with Text::ParseWords 3.29 bundled in Perl 5.16.3.

ikedas:
@xandm , how about [my proposal](https://github.com/sympa-community/sympa/issues/466#issuecomment-628930642)? I plan to submit another PR in some weeks.

ikedas:
This issue is merged into #466 by now.

PR #941, opened 2020-05-18
Every time users were loaded via include_sql_query Sympa would raise warnings

salaun-urennes1:
Obviously the issue was introduced back in 2015, as mentionned in https://listes.renater.fr/sympa/arc/sympa-users/2015-12/msg00002.html.

Here is an example of the issue :
```
/usr/local/sympa/bin/sympa.pl --sync_include=mylist-with-included-members@listes.etudiant.univ-rennes1.fr
...
Use of uninitialized value $value in printf at /usr/local/sympa/bin/Sympa/DataSource/SQL.pm line 91, <GEN1> line 1.
Use of uninitialized value $value in printf at /usr/local/sympa/bin/Sympa/DataSource/SQL.pm line 91, <GEN1> line 1.
Use of uninitialized value $value in printf at /usr/local/sympa/bin/Sympa/DataSource/SQL.pm line 91, <GEN1> line 1.
include [notice] Subscribers were included from data source "mysource" into list mylist-with-included-members.
include [notice] Inclusion succeeded (69 added, 15 deleted, 2175 updated).
```

### DIFF
--- a/src/lib/Sympa/DataSource/SQL.pm
+++ b/src/lib/Sympa/DataSource/SQL.pm
@@ -88,7 +88,7 @@ sub _open {
                     $email =~ s/[\t\r\n]+/ /g;
                     $value =~ s/[\t\r\n]+/ /g if defined $value;
 
-                    printf $tmpfh "%s\t%s\n", $email, $value;
+                    printf $tmpfh "%s\t%s\n", $email, $value // '';
                 }
             }
             $sth->finish;
--- a/src/lib/Sympa/DataSource/SQL.pm
+++ b/src/lib/Sympa/DataSource/SQL.pm
@@ -86,14 +86,9 @@ sub _open {
                     my ($email, $value) = @$row;
                     next unless defined $email and length $email;
                     $email =~ s/[\t\r\n]+/ /g;
-					
-					# The SQL query may return the user name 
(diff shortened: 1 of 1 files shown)


racke:
Why not:

 `printf $tmpfh "%s\t%s\n", $email, defined $value ? $value : '';`

It might need parenthesis, but it is shorter and with correct indentation.

salaun-urennes1:
It's probably a matter of taste, but I'm in favor of code readability and I don't find the single line version easy to read. But feel free to patch the code that way, no problem.

Regarding indentation, I think it's a problem with my Vim setup, sorry about that. I'll try to fix that...

racke:
> It's probably a matter of taste, but I'm in favor of code readability and I don't find the single line version easy to read. But feel free to patch the code that way, no problem.

Yeah fair enough. 

> 
> Regarding indentation, I think it's a problem with my Vim setup, sorry about that. I'll try to fix that...

Tip: Use  "git diff --cached" before commit or "git show" before push. 

At any rate, I definitely appreciated the fix :+1:

ikedas:
Now we support Perl 5.10.1 or later and may also do
``` perl
printf $tmpfh "%s\t%s\n", $email, $value // '';
```

PR #955, opened 2020-06-07
Deprecate implicit sync_include in Sympa::List constructor

### LINKED ISSUE #92: (Re)allow lists to only have owners from data sources
This might be more of a feature request then a bug rapport. 
In version 5 it was possible to have lists that only had owners from a data source. There was no need to set one manually a well.

If you try to remove the last owner in 6.2.16 you get the following error message.

`ERROR (edit_list) - Parameter 'owner' is mandatory. Ignoring deletion. 
ERROR (edit_list) - No parameter was edited `

ikedas:
- review: Won't process sync_include in advance so that processing will be faster.
    sync_include just before message delivery will be performed as in the past.

  - sync_include will no longer be invoked during Sympa::List->new():
      - No owner defined in a list is no longer treated as error_config.
      - Instead, if no owner defined and something has to be sent to owners,
          - If possible, discard incoming message and send back DSN to original sender,
          - or, notifications to owners will be redirected to listmaster(s).

This may improve speed of "review" page and so on.
Also, this may fix #92 .

### DIFF
--- a/default/Makefile.am
+++ b/default/Makefile.am
@@ -383,6 +383,7 @@ nobase_default_DATA = \
 	mail_tt2/request_auth.tt2 \
 	mail_tt2/report.tt2 \
 	mail_tt2/review.tt2 \
+	mail_tt2/shared_moderate.tt2 \
 	mail_tt2/send_auth.tt2 \
 	mail_tt2/sendpasswd.tt2 \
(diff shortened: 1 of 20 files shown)


ikedas:
@salaun-urennes1 , could you please check this change is useful?

PR #968, opened 2020-07-04
Use canonical and current link in the description of the password_validation config parameter

racke:


ikedas:
@racke , could you please work on [translation site](https://translate.sympa.org/en_US/sympa/translate/en_US.po#unit=58870) ?

Issue #993, opened 2020-09-08
Rebuilding all archives should remember where it stopped


ldidry:
Version
-------
Sympa 6.2.54, Debian Buster

Installation method
-------------------
Source package

Expected behavior
-----------------
If the `archived.pl` dæmon restart for a reason or another (like crash, logrotate…) while rebuilding all lists, it should not restart rebuilding all lists from the beginning but from the last rebuilt list.

Actual behavior
---------------
When the `archived.pl` dæmon restart for a reason or another (like crash, logrotate…) while rebuilding all lists, it  restarts rebuilding all lists from the beginning.

Additional information
----------------------
We have more than 37,000 lists, we need to rebuild the archives to remove the attachments from the archives (changes of terms of use) and unfortunately, the `archived.pl` does not stay up long enough to rebuild all archives. So it rebuilds, again and again (and unfortunately, the rebuilding locks the new messages archiving).

PR #994, opened 2020-09-08
sympa.pl: Add an option "--rebuildarc=list@domain" to rebuild archives

ldidry:
Relative to #993: with this option, we can rebuild all archives with a simple `for` bash loop, without restarting from beginning if `archived.pl` restarts while rebuilding all lists.

### DIFF
--- a/src/sbin/sympa.pl.in
+++ b/src/sbin/sympa.pl.in
@@ -48,8 +48,10 @@ use Sympa::Language;
 use Sympa::List;
 use Sympa::Log;
 use Sympa::Mailer;
+use Sympa::Message;
 use Sympa::Spindle::ProcessDigest;
 use Sympa::Spindle::ProcessRequest;
+use Sympa::Spool::Archive;
 use Sympa::Template;
 use Sympa::Tools::Data;
 use Sympa::Upgrade;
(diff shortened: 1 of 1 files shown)


ikedas (commented):
The keyword `rebuild_arc` has never been used.  I feel `rebuildarc` looks less confusing.

ldidry:
I have no pb with that. I choosed `rebuild_arc` since it’s closer to natural language.

racke:
Natural language would be rebuild_archives. An arc is part of a building.

ldidry:
I said "closer", not "exactly" :sweat_smile:

ldidry:
Now, it’s `rebuildarc`.

dverdin (approved):
It's a quick and useful utility. I see absolutely nor problem with it.
Actually, I wonder whether the "rebuild all" command should be declined in a similar way : a set of single "rebuild list" tasks for each list.

(in the doc, I would say "rebuild the list's archives" but then again it's perfectly understandable this way)

ldidry:
> Actually, I wonder whether the "rebuild all" command should be declined in a similar way : a set of single "rebuild list" tasks for each list.

It may be more simple than track the state of "rebuild all lists" task.

While it’s doable, it could take a lot of time to create all the "rebuild list" messages if there is many lists, so it would make a timeout on the web interface. I suggest, if this is a thing we want to do, to create a "schedule rebuild all list" message that would create all the "rebuild list" messages when processed.

What do you think?

dverdin:
Yes, that's certainly the best approach.
This way, the request is instantaneous.
However, it's way out of the current PR's scope. It's perfect this way.

ikedas (approved):
I found no problem.

ikedas:
Merged for now.  Please check the next beta in tomorrow.

PR #1004, opened 2020-09-22
Update a dependency MHonArc

ikedas:
See also sympa-community/MHonArc#7

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -96,7 +96,7 @@ requires 'List::Util::XS', '>= 1.20';
 requires 'Locale::Messages', '>= 1.20';
 
 # MHonArc is used to build Sympa web archives
-requires 'MHonArc::UTF8';
+requires 'MHonArc::UTF8', '>= 2.6.23';
 
 # Required to compute digest for password and emails
 requires 'MIME::Base64', '>= 3.03';
--- a/cpanfile
+++ b/cpanfile
@@ -96,7 +96,7 @@ requires 'List::Util::XS', '>= 1.20';
 requires 'Locale::Messages', '>= 1.20';
 
 # MHonArc is used to build Sympa web archives
-requires 'MHonArc::UTF8', '>= 2.6.23';
+requires 'MHonArc::UTF8', '>= 2.6.24';
 
 # Required to compute digest for password and emails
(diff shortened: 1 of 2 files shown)


ikedas:
This fix would be better to be released when the packagers will have been prelared new packages of MHonArc.

racke:
MHonArc is not yet available on CPAN.

ldidry:
And the version currently on PAUSE does not have MHonArc.pm

racke:
> And the version currently on PAUSE does not have MHonArc.pm

@ldidry so how to we go forward, flush the version form PAUSE which seems to be stuck anyway and prepare one with MHonArc.pm?

ldidry:
I think that it’s indeed better to remove the current version from PAUSE (I juste scheduled the deletion) in order to re-add the MHonArc.pm.
I’ll push a branch on github tomorrow to see if its content seems good to you.

ikedas:
> MHonArc is not yet available on CPAN.

See https://metacpan.org/release/LDIDRY/MHonArc-2.6.22
However it has not yet been indexed for cpanm:
http://cpanmetadb.plackperl.org/v1.0/package/MHonArc::UTF8

ldidry:
Should I remove the scheduled deletion of the file?

racke:
> Should I remove the scheduled deletion of the file?

I don't think so. While the direct link above works, it is not indexed within MetaCPAN. What did the email from PAUSE tell you about the indexing?

ldidry:
It failed when uploading the archive. A day or two after that, I got a mail from Neil Bowers.

racke:
That's a tough nut. Can't understand why PAUSE doesn't clean up broken packages by itself. 
At any rate, let's get up a release with MHonArc.pm.

ikedas:
@ldidry , is there any progress?

ldidry:
I just uploaded a new version (2.6.23) to CPAN

racke:
Great, thank you!

racke:
Commandline usage seems to be broken though: https://github.com/sympa-community/MHonArc/issues/12

ldidry:
Wait a bit for the 2.6.24 to be indexed on PAUSE and update the dependency to 2.6.24.

Issue #1009, opened 2020-10-07
[CVE-2020-26880] root privilege escalation from user sympa by modifying sympa.conf


Beuc:
**Note by admin: this issue is dedicated to measures against  [CVE-2020-26880](https://nvd.nist.gov/vuln/detail/CVE-2020-26880).**

---

Version
-------
any

Installation method
-------------------
any

Expected behavior
-----------------
user 'sympa' only execute specific tasks as root and cannot gain more privileges, following the principles of privileges separation

Actual behavior
---------------
user 'sympa' can obtain full root shell access

Additional information
----------------------

Following up on https://github.com/sympa-community/sympa/issues/943#issuecomment-704779420
> A more concrete vulnerability you document in your patch, is that sympa.conf (as well as its entire directory) is owned by user 'sympa', so an attacker who compromised [the 'sympa' user] can directly edit the main [sympa.conf] and escalate to root.

Given that the configuration file is parsed as root through the setuid sympa_newaliases-wrapper, and can execute arbitrary commands through its backticks syntax, there is an unintentional privilege escalation from sympa to full root shell access.

The information is public in the aforementioned thread hence this bug introduce no confidential information and is reported publicly as well.
(This also allows for CVE assignment for small projects without a CNA able to run a full embargo procedure.)

Mitigations include replacing sympa_newaliases-wrapper by an alternate alias manager, see e.g. https://tribut.de/blog/sympa-and-postfix .

Credits goes to @lightsey .

ikedas:
Hi @Beuc ,
Please e-mail me. I want to talk with you privately.

Beuc:
To clarify, the backticks syntax is the most trivial method, but probably not the only one.

The core issue identified here, is running sympa as root with a configuration file modified by the attacker.

ikedas:
> To clarify, the backticks syntax is the most trivial method, but probably not the only one.
> 
> The core issue identified here, is running sympa as root with a configuration file modified by the attacker.

Which is the core issue, either that Sympa is running as root or that a configuration file is modified by the attacker?

Beuc:
This was assigned CVE-2020-26880.
(It tracks the combination of both.)

I agree with your understanding.
For clarity here, what I suggested in my mail is to only run /usr/bin/newaliases as root (rather than sympa itself).

[...]

PR #1015, opened 2020-10-12
Add "aliases_wrapper" configuration parameter

### LINKED ISSUE #946: Use alias wrapper only if it is really needed
This is a follow up to #943. It is established security practice to only elevate privileges if really needed. Currently we always call the wrapper instead of the newaliases program, which is a bad idea.

Especially as the wrapper is usually not needed. With postfix you are usually running postmap. Exim can use plain text files, so it doesn't need any newaliases program.

So we could call the program directly (at least in the default configuration).

Does that make sense?

racke:
This allows you to invoke the command from "sendmail_aliases" without using the wrapper. This should fix #946 and should work for Exim and Postfix MTA in most cases.

Current default is to use the alias wrapper, but in the future we may change that default.

### DIFF
--- a/src/lib/Sympa/Aliases/Template.pm
+++ b/src/lib/Sympa/Aliases/Template.pm
@@ -39,8 +39,6 @@ use base qw(Sympa::Aliases::CheckSMTP);
 
 my $language = Sympa::Language->instance;
 my $log      = Sympa::Log->instance;
-my $alias_wrapper =
-    Sympa::Constants::LIBEXECDIR . '/sympa_newaliases-wrapper';
 
 sub _aliases {
     my $self = shift;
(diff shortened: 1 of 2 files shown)


ikedas (approved):
There no longer are nothing to review for me.

Beuc:
Hi, what is the status of this pull request?
Minor precision for clarity, I believe the command referenced in OP is 'aliases_program', or 'sympa_newaliases.pl' (not 'sendmail_aliases').

racke:
> Hi, what is the status of this pull request?

It is supposed to be ready for the next release.

> Minor precision for clarity, I believe the command referenced in OP is 'aliases_program', or 'sympa_newaliases.pl' (not 'sendmail_aliases').

Sorry I miss the clarity in your comment, please elaborate.

mhow2:
@ikedas : it seems Sylvain is waiting for an answer at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=972114

This fix is still pending in debian, as far as I understand. (still 1 ignored security issue at https://tracker.debian.org/pkg/sympa)

Beuc:
@mhow2 this issue is pretty much independent from CVE-2020-26880, see #1009 instead.
Also I added a comment at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=972114#29 to clarify.

racke:
Yes, in my opinion no part of Sympa should be run as another user. This would fix root escalation issues and it is feasible as far I can tell. But we need more documentation about configuring Sympa which each MTA and webserver to reach this goal.

Issue #1018, opened 2020-10-13
Mail loop for sympa-request address


StianAndreassen79:
Hi Sympa Team,

Thanks for a great product! We have used it for years here at James Cook University. We have recently upgraded from 6.1.4 to 6.2.56, and we're seeing an issue that others have reported with this version. It's pertaining to how messages to sympa-request appear to be delivered to the sympa queue, but rejected as there's no such list, and they never should have been routed there in the first place.

Other bug reports have been logged regarding this;
https://github.com/sympa-community/sympa/issues/995
https://github.com/sympa-community/sympa/issues/957

We saw the same issue on our Sympa once we opened the new production server to mail in and out, it never happened in our test environment that was run for a while before we cut over to the production server. The issue manifested itself in extremely verbose log files, overworked spools and eventually RAM issues leading to paging and RAM initialization issues for the main Sympa threads. The delivery attempt occurred several times pr minute, the worst case I found was 143 times pr minute, all day and all night.

We implemented a workaround of adding a regexp to our virtual.regexp file, which is a virtual_maps regex defined in main.cf, as we don't use sendmail aliases, we use only regexp in our postfix configuration. This line effectively forwards these messages to /dev/null and the issue disappears;
`/^sympa-request\@<list-domain>\.<dc>\.<dc>\.<dc>$/ devnull@localhost`
We then have an entry in /etc/aliases that handles this; 
devnull: /dev/null
as indicated in the other issues logged about this.

The reason why I'm submitting this issue, is that it would appear that the other issues logged seem to point to the root cause of this issue is a postfix misconfiguration where there are emails being routed IN to the Sympa host, sent to "sympa@listdomain.maildomain", evidenced by Sympa's log entries. But how can it be a postfix issue, if the 
"FROM" address entries in both Sympa and postfix's log files show that;
1. maillog has NO value for "from"
2. sympa.log has "recipients=sympa-request@listdomain.maildomain; sender=sympa-request@listdomain.maildomain"

This leads me to believe that the emails appearing to be sent to sympa@listdomain.maildomain are actually being sent by Sympa itself, or may be a result of only using regexp configuration for postfix? Our sympa_aliases file is identical to what's in the documentation, but obviously having the list and mail domains modified. What are the chances you could look at this issue again?

ikedas:
Hi @StianAndreassen79 ,
Could you please show a concrete example of log lines in which there is 'NO value for "from"'?  I cannot understand what you mean.

[...]

PR #1019, opened 2020-10-14
Workaround for mail loop for sympa-request address (#1018)

ikedas:
As of Postfix 2.3, Content of Return-Path: pseudo-header field added by pipe(8) for null envelope sender became "MAILER-DAEMON".  Empty `null_sender` option will prevent this behavior.
As a workaround, the uppercase "<MAILER-DAEMON>" will be treated as null envelope sender.

This is a workaround for #1018 .

### DIFF
--- a/src/lib/Sympa/Message.pm
+++ b/src/lib/Sympa/Message.pm
@@ -134,6 +134,9 @@ sub new {
         my $addr = $1;
         if ($addr =~ /<>/) {    # special: null envelope sender
             $self->{'envelope_sender'} = '<>';
+        } elsif ($addr =~ /<MAILER-DAEMON>/) {
+            # Same as above, but a workaround for pipe(8) of Postfix 2.3+.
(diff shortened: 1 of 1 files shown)


src/lib/Sympa/Message.pm
```diff
@@ -4377,6 +4380,10 @@ Prepending C<Return-Path:> is available by default.
 
 Add C<R> to the C<flags=> attributes in master.cf.
 
+Additionally with Postfix 2.3 or later, add an empty C<null_sender=>
+attribute.
+Or "null envelope sender" would be replaced with C<E<lt>MAILER-DAEMONE<gt>>.
```
racke, line 4385: Typo: s/MAILER-DAEMONE/MAILER-DAEMON/.

ikedas: `E` is a part of `E<gt>`.

Issue #634, opened 2019-06-03
Add « report as spam » button in archives


ldidry:
Expected Behavior
-----------------
Currently, only mails awaiting for moderation can be reported as spam

Current Behavior
----------------
I wish moderators/owners to be able to report mail as spam even if it has been already delivered (use case: non-moderated lists).

This would make the antispam able to learn spams that it missed.

Possible Solution
-----------------
Add « report as spam » button in archives pages for moderators/owners

Context
-------
Public and open to anyone Sympa instance (framalistes) with many non-moderated lists.

PR #1022, opened 2020-10-14
🐛 — Fix confirmation for reporting as spam while deleting an archived mail

ldidry:
Currently, if you want to report archived mail(s) as spam while asking for deletion, the confirmation page is shown twice (only once if not reporting as spam). It appears like a bug to users.
This commit remove the confirmation when reporting as spam (there is still the confirmation for the deletion).

### DIFF
--- a/src/cgi/wwsympa.fcgi.in
+++ b/src/cgi/wwsympa.fcgi.in
@@ -8978,11 +8978,8 @@ sub do_remove_arc {
 
     # Action confirmed?
     my $next_action = $session->confirm_action(
-        $in{'action'},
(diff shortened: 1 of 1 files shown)


ldidry:
Relative to #634 and #640

ikedas (approved):
I feel it proper.  I agree to add it to the next milestone, 6.2.60.

dverdin (approved):
Simple and efficient. I agree!

ikedas:
:beer:

ldidry:
Since 6.2.58 has been released, shouldn’t it be merged?

PR #1024, opened 2020-10-14
Renaming of lists which members are included in other lists 

salaun-urennes1:
This was foridden. This commit changes this behavior. 

After renaming is done, the new code goes through ->get_including_lists('member') and updates config of these lists to include members with the new listname. Note however that inclusions through data_sources (member_include, owner_include or editor_include parameters) are not updated; too hazardious to perform source_parameters changes without knowing the semantics.

### DIFF
--- a/default/web_tt2/admin.tt2
+++ b/default/web_tt2/admin.tt2
@@ -56,7 +56,7 @@
             </form>
         [%~ ELSIF is_included ~%]
             <p>
-                [%|loc%]Closing or renaming this list is impossible, because it is included by other mailing list(s).[%END%]
+                [%|loc%]This list is included by other lists and cannot be closed.[%END%]
                 <br />
                 <a class="button" href="[% 'including_lists' | url_rel([list]) %]">
                     [%|loc(list)%]View lists including %1[%END%]
@@ -71,11 +71,11 @@
                 </div>
             </form>
         [%~ END %]
-        [% IF may_create_list && ! is_included ~%]
+        [% IF may_create_list ~%]
             <form name="manage_list_name" action="[% path_cgi %]" method="post">
                 <fieldset>
                     <input class="MainMenuLinks" type="submit" name="action_rename_list_request" value="[%|loc%]Rename List[%END%]"/>
-                    [%|loc%]Allows you to change this list's name. Everything related to the list will be renamed, including the mail aliases and the web archives.[%END%]
+                    [%|loc%]Allows you to change this list's name. Everything related to the list will be renamed, including the mail aliases and the web archives. If members of this list are included as members in other lists, these inclusions will be updates too. However inclusions through data_sources cannot be treated.[%END%]
                     <input type="hidden" name="list" value="[% list %]"/>
                 </fieldset>
             </form>
--- a/src/lib/Sympa/Request/Handler/move_list.pm
+++ b/src/lib/Sympa/Request/Handler/move_list.pm
@@ -84,21 +84,15 @@ sub _twist {
         return undef;
     }
 
-    # If list is included by another list, then it cannot be renamed.
+    # If list is included by another list, then we need to change these lists config
+    # First step: keep an array of these lists
+    # Second step (after list renaming): update these lists config
+    my $members_included_by = [];
     unless ($mode and $mode eq 'copy') {
         if ($current_list->is_included) {
-            $log->syslog('err',
-                'List %s is included by other list: cannot rename it',
-                $current_list);
-            $self->add_stash(
-                $request, 'user',
-                'unable_to_rename_list',
-                {   listname     => $current_list->get_id,
-                    new_listname => $listname . '@' . $robot_id,
-                    reason       => 'included'
-                }
-            );
-            return undef;
+            # We don't consider members included as owner/editor in other lists
+            # This would be tricky because these inclusions use third party data_sources 
+            $members_included_by = $current_list->get_including_lists('member');
         }
     }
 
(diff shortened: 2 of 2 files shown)


ikedas:
@salaun-urennes1 , what is your motivation?

salaun-urennes1:
We have 10 robots on our production server. We need to move 120 mailing lists from one robot to another for organizational reasons. Most of these 120 mailing lists have inclusion dependencies. I it stands the code does not even let me move the included lists to another robot. I need to move the lists and make sure inclusion relationships are updated too.

ikedas:
I don't recommend to change Sympa for the purpose you described.

Renaming operation will imply updating large number of lists, while it may be done through the web interface.  If the process of renaming was aborted by any reason, list configuration might become inconsistent.  And I think that's why renaming the list that has included users is prohibited.

You may be able to write batch script to change inclusion setting of single list, then run it by each list.

salaun-urennes1:
I agree that renaming via the web interface has the risk of leaving inconsistencies because the processing is slow (especially if the renamed list is included by many lists) and it could leed to timeouts.

For my use case I did a batch processing running `sympa.pl --rename_list` with the new code. 

I still think sympa native code should provide that capability of renaming included lists. Therefore I propose to 1) keep the changes in move_list.pm (so that `sympa.pl --rename_list` can do the job) 2) restore the original admin.tt2 web template and complete it to suggest using the CLI to rename the list.

ikedas:
I think you'd be better to devide the renaming process into small pieces, instead of implementing complex large function.

If I performed such renaming, I would do it as following:

  1. Copy the list `source@old.domain` to the new list `source@new.domain` (not renaming but duplicating the list).  Also copy all of the users (member, owner, editor) to the new list.
  2. [During this step delivery to the list should be stopped]
       - Edit aliases file to forward the address of the old list to those of the new list so that the message to the both lists will be delivered to the new list.
       - Move archives of the old list to the new list (If this will take a long time, only the last month of archives may be moved. 
 The rest may be moved afterward).
  3. By each list which are including `source` list, edit config to include from the new list.
  4. Confirm that included users are updated correctly.
  5. Remove the old list `source@old.domain`
 
I think this is safer, and don't need implementing new feature.

salaun-urennes1:
The steps you describe are already implemented by Sympa during rename_list; no need to rewrite that process in a homemade script IMHO. 

Given the process is done using `sympa.pl --rename_list` (and not from the web interface), can you explain why you consider this unsafe?

ikedas:
> The steps you describe are already implemented by Sympa during rename_list

Do you mean the `move_list` request handler? If you do, the steps I describe have not implemented yet.  Please compare actual code and my description.

salaun-urennes1:
Yes I was referring to  move_list request handler. 
The code does the following actions:
1. move the list directory,
2. update all database entries (subscriber_table, admin_table, list_table, exclusion_table, inclusion_table, stat_table, stat_counter_table),
3. update spools,
4. rename archives,
5. take care of mail aliases

And this PR takes care of inclusions.

In what circumstances do you consider the process unsafe? 
What part of the code should be improved?

ikedas:
> 3. By each list which is including source list, edit config to include from the new list.
> 4. Confirm that included users are updated correctly.

I think these steps is more complex than you think: For example, on step 3, `$other_list->save_config()` can't always update inclusion settings.  That's why I suggest step 4: It's dangerous not to confirm updated config is properly working --- This step cannot be automated.

salaun-urennes1:
I don't intend to make this thread too long, but can you explain why `$other_list->save_config()` can't always update inclusion settings?

ikedas:
> I don't intend to make this thread too long, but can you explain why `$other_list->save_config()` can't always update inclusion settings?

You don't consider member_include parameter.  It refers *.incl file that is parsed as template.  Thus, this parameter cannot be updated just by replacing list name.

There may be the other differences.  One I realized for now is that --- My suggestion keeps the old list until all steps have been finished.  Your code removes it before inclusion settings of all lists have been updated.  What would happen if the sync_include task run at this time?

salaun-urennes1:
> You don't consider member_include parameter. It refers *.incl file that is parsed as template. Thus, this parameter cannot be updated just by replacing list name.
That's a limitation I clearly documented in the proposed code and also in admin.tt2.

> There may be the other differences. One I realized for now is that --- My suggestion keeps the old list until all steps have been finished. Your code removes it before inclusion settings of all lists have been updated. What would happen if the sync_include task run at this time?
If we do it the other way around (update lists that include the current one before the list is renamed) it's much more problematic because 1) the new listname does not exist yet and 2) list renaming process could fail, thus leaving inconsistent inclusions. 
The alternative you propose (writing my own script to cope with the renaming) brings the same risk of task_manager running a sync_include task during the renaming process. I can think of 2 ways to prevent that: 1) add a warning when `sympa.pl --rename_list` is run asking the admin to stop task_manager process or 2) disable that list's tasks in `move_list()`. I can complete the PR in that direction if you wish...

ikedas:
> > You don't consider member_include parameter. It refers *.incl file that is parsed as template. Thus, this parameter cannot be updated just by replacing list name.
> 
> That's a limitation I clearly documented in the proposed code and also in admin.tt2.

Don't you know that the users often won't read the fancy manual? :-/

> > There may be the other differences. One I realized for now is that --- My suggestion keeps the old list until all steps have been finished. Your code removes it before inclusion settings of all lists have been updated. What would happen if the sync_include task run at this time?
> 
> If we do it the other way around (update lists that include the current one before the list is renamed) it's much more problematic because 1) the new listname does not exist yet and 2) list renaming process could fail, thus leaving inconsistent inclusions.
> The alternative you propose (writing my own script to cope with the renaming) brings the same risk of task_manager running a sync_include task during the renaming process. I can think of 2 ways to prevent that: 1) add a warning when `sympa.pl --rename_list` is run asking the admin to stop task_manager process or 2) disable that list's tasks in `move_list()`. I can complete the PR in that direction if you wish...

If you wish to achieve your goal, renaming of lists which members are included in other lists, you would be better to put the improvement of Sympa on the back burner and to make a plan to carry it out _manually_.

[My suggestion above](https://github.com/sympa-community/sympa/pull/1024#issuecomment-709703938) is written considering manual operation.  It may have some check points from where you can turn back, flexiblly coping with various situations.

How about withdrawing this PR for the present?

salaun-urennes1:
Too bad this can't be included in Sympa code.

PR #1025, opened 2020-10-26
Show subjects of archived mails before deletion’s confirmation

ldidry:
I think it would be good to have a reminder of which archived mails are about to be deleted (just like the reminder of the subscribers who are about to be unsubscribed), so here it is.

### DIFF
--- a/default/mhonarc-ressources.tt2
+++ b/default/mhonarc-ressources.tt2
@@ -235,7 +235,12 @@ IDXPGEND
 <LITEMPLATE>
         <li>
             ($tag$% IF is_owner %$tag$)
-                <input type="checkbox" name="msgid" value="$MSGID$" />
+                ($tag$% msgidx = msgidx || 0; msgidx = msgidx + 1 ~%$tag$)
+                <input type="checkbox" name="msgid" value="$MSGID$"
+                 id="msgid($tag$% msgidx %$tag$)"/>
+                <input type="hidden" name="msg_subject" value="$SUBJECTNA$"
+                 class="disableUnlessChecked"
+                 data-selector="#msgid($tag$% msgidx %$tag$)" />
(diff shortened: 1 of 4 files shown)


ikedas (commented):
``` html
                <input type="checkbox" name="msgid" value="$MSGID$_SYMPA_$SUBJECTNA$" />
```
This will make HTML archives not backward-compatible.  (And, this is a matter of personal taste, but separator `_SYMPA_` looks a bit inelegant for me).

Instead of doing as above, for example, we may use jQuery to enable/disable `msg_subject`, doing such as:
``` html
                ($tag$% msgidx = msgidx || 0; msgidx = msgidx + 1 ~%$tag$)
                <input type="checkbox" name="msgid" value="$MSGID$"
                 id="msgid($tag$% msgidx %$tag$)" />
                <input type="hidden" name="msg_subject" value="$SUBJECTNA$"
                 class="disableUnlessChecked"
                 data-selector="#msgid($tag$% msgidx %$tag$)" />
```
About `disableUnlessChecked` class see `www/js/sympa.js`.

ldidry:
I thought about using javascript but I wanted to have a solution for those who don’t use JS.
But if we already have a thing for this kind of case, I can use it.

Indeed, it’s not backward compatible but we can have a message in the CHANGELOG saying that admins can rebuild archives… or not, since it doesn’t break anything for old archives, it just have the old behavior :slightly_smiling_face: 

I’ll recode it to use JS

ldidry:
Here we go: new method, with JS.

ikedas (approved):
I found no problem. Thanks for improvement!

ikedas:
:beers:

ldidry:
Since 6.2.58 has been released, shouldn’t it be merged?

PR #1032, opened 2020-11-12
Add option to ./configure which prevents installation of sympa_newaliases-wrapper (#1031)

### LINKED ISSUE #1031: Add option to ./configure which prevents installation of sympa_newaliases-wrapper
The setuid wrapper is not needed for Exim or Postfix with the new `aliases_wrapper` configuration option from #1015. I suppose it isn't necessary for Sendmail as well.

In my opinion there is no point to produce a binary which poses a security risk in this case. If you agree with me, I'm going to produce a corresponding pull request.

Recommended configuration for Exim:

`
aliases_program: none
`

Recommended configuration for Postfix (single domain):

```
aliases_program: postalias
aliases_wrapper: off
```

Recommended configuration for Postfix (multiple domains):

```
aliases_program: postmap
aliases_wrapper: off
```

ikedas:
configure: Add options --disable-setuid_newaliases and --disable-setuid

This may fix #1031

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -608,6 +608,17 @@ AC_ARG_WITH(
 )
 AC_SUBST(POSTMAP_ARG)
 
+AC_MSG_CHECKING([whether we install/enable setuid wrappers])
+AC_ARG_ENABLE(
+    setuid,
+    AS_HELP_STRING(
+        [--disable-setuid],
+        [do not install nor enable any setuid wrappers]
+    ), ,
+    [enable_setuid="yes"],
+)
+AC_MSG_RESULT([$enable_setuid])
+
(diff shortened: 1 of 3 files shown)


racke:
Looks good to me. I will test it in my development environment.

Cosmetics: can we use --disable-setuid-newaliases instead of --disable-setuid_newaliases?  I'm not a friend of mixing dashes and underscore.

ikedas:
> Cosmetics: can we use --disable-setuid-newaliases instead of --disable-setuid_newaliases? I'm not a friend of mixing dashes and underscore.

I'm not familiar with it, too, but it looks a custom of Sympa, e.g. `--with-aliases_file`.

Issue #501, opened 2018-12-11
Raw parameters in the footer


mathieuroquefort:
Hello,

I got complaints from our users because the links in the footer didn't work.
And it appears that the parameters are not replaced by their values.

Here is the footer in the messages :
```
Si vous désirez vous désabonner, envoyez un message en cliquant sur ce lien : mailto:sympa@domain.tld?subject=uns%20[% list.name %]
. 
Si vous désirez modifier votre abonnement, rendez-vous à cette adresse : 
[% conf.wwsympa_url %]/sympa/suboptions/[% list.name %] 
```

It did work previously. I don't understand what happened...

Version
-------
Sympa 6.1.17 / CentOS 6.10

Installation method
-------------------
?

Expected behavior
-----------------
[% list.name %] should be replace by the list name, and [% conf.wwsympa_url %] by the URL of the web interface.

Actual behavior
---------------
Parameters are not replaced, and displayed raw.

Additional information
----------------------

ikedas:
Hi @mathieuroquefort,

From list info page, please go to [Edit list config] - [Sending/receiving setup], and check if [Attachment type] is "append to message body".  If "add a new MIME part" is set, footer will be appended as a separate part and won't be replaced.

mathieuroquefort:
I've just tried it. Now, on the list parameters, footer_type = append

But then :
- If I send an html message, no footer is added
- If I send a plain text message, the footer is appended but the parameters are not replaced

I had already tested different parameters here before.
The other problem is, that almost nobody sends plain text message now. I don't know how to append footer on html messages, and that is a real problem.

ikedas:
I realized that you use a bit earlier version of Sympa.  Please upgrade to recent version --- if possible, 6.2.x, or at least 6.1.25 --- and check if problem will be solved.

mathieuroquefort:
We can't do that easily : we use a managed server, running CentOS. And CentOS hasn't been tested with sympa 6.2.
We use, I think, the latest validated version with CentOS6.10. Our manager doesn't like using non-validated softwares.
Therefore, I'm looking for a more short-term solution. For now on, I have replaced the parameters with their values on each list. With 300 lists that's quite a job, but that was the only solution.

Does the 6.2 version of sympa improve the footer appending in html messages ? I can't find this information in the documentation.

ikedas:
I don't know what is "validation with ContOS" (Sympa 6.1 also have not been validated. Besides).

[...]

PR #1038, opened 2020-11-19
Limited personalization

### LINKED ISSUE #1037: Personalization (merge_feature) should be limited
Version
-------

6.2.58 and earlier

Expected behavior
-----------------

  - If personalization is enabled, the messages posted only by limited users should be allowed personalization.
  - Or, scope of personalization should be limited to where owners can control: Only footer/header.

Actual behavior
---------------

  - Once personalization is enabled, any users who were allowed posting can use this feature.
  - And, entire text body of the message can be in the scope of conversion.

ikedas:
Limit personalization:

- Only when posted from WWSympa, entire text will be converted. Otherwise only footer/header will be converted.

This may fix #1037 .

### DIFF
--- a/default/edit_list.conf
+++ b/default/edit_list.conf
@@ -156,6 +156,8 @@ max_list_members		editor				hidden
 message_hook			owner				hidden
 message_hook			privileged_owner		read
 
+personalization			owner,privileged_owner		read
+
 default 			privileged_owner 		write
 default 			owner 				write
 default 			editor 				read
--- a/default/web_tt2/compose_mail.tt2
+++ b/default/web_tt2/compose_mail.tt2
@@ -53,7 +53,7 @@
             <textarea name="body" id="body" cols="80" rows="25">[% body %]</textarea>
             <br />
 
-            [% IF merge_feature %]
+            [% IF listconf.merge_feature == 'on' %]
                 <br />
                 <b>[%|loc%]Messages customization: use the template syntax:[%END%] <a href="http://www.tt2.org">TT2</a></b>
                 <br />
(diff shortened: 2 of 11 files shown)


racke:
I agree with you that personalization should be restricted. However, I have existing projects where personalized emails are sent by email to Sympa. I suggest to control this by scenari.

ikedas:
racke, thanks for comment.

Using scenario seems worth considering.
Another idea I came up with at first is subdividing the value of `merge_feature` parameter, e.g. `full`, `footer` and `none` instead of `on` & `off`.

racke:
More detailed configuration makes a lot of sense to me. I suggest that we add an additional configuration parameter which comes with `merge_feature` instead of subdividing it.

For example `merge_parts` with values like full, footer, body.

As everything else is controlled by scenari, it would make sense to add one for the merge_feature too.

ikedas:
> More detailed configuration makes a lot of sense to me. I suggest that we add an additional configuration parameter which comes with `merge_feature` instead of subdividing it.
> 
> For example `merge_parts` with values like full, footer, body.
> 
> As everything else is controlled by scenari, it would make sense to add one for the merge_feature too.

Ok, I'll think of adding new parameter and scenario.

ikedas:
@racke, I added new parameters they may take a value either of `none`, `footer` (only header and footer) or `all` (header, footer and body).
  - `personalization.web_apply_on`: By default `footer`.
  - `personalization.mail_apply_on`: By default `none`.

Administrators may have to edit configuration file, if they want to re-enable potentially insecure behavior.

ikedas:
We should test if this pr will change behavior reported in #501 .

PR #1044, opened 2020-11-27
Properly check email and session id in authenticateAndRun SOAP call (#1041)

### LINKED ISSUE #1041: Unauthorised full access via SOAP API due to illegal cookie
Version
-------
v6.2.56-1.el7 on Centos 7.8.2003

Installation method
-------------------
Centos package

Expected behavior
-----------------
permission denied

Actual behavior
---------------
error message and action actually executed anyways.

Additional information
----------------------
In our setup we have a problem with incorrect cookies via the SOAP API of sympa.
If the SOAP request contains a correct cookie everything works as expected -> request executed
If the SOAP request contains a correct but outdated cookie, everything works as expected -> request correctly denied.

If the SOAP request contains an arbitrary string as cookie (e.g. "asdkjasdljkahsdlkjh"), SOAP replies with an error ("Undefined session ID in cookie") but STILL executes every requests we make. By this we can add email adresses to lists without authentication, any operation we tried was still successful. 

We could hotfix the problem by inserting a die(); command into /usr/share/sympa/lib/Sympa/WWW/Session.pm:129 like this:
```
    my $session_id = _cookie2id($cookie);
    unless ($session_id) {
        $log->syslog('info', 'Undefined session ID in cookie "%s"', $cookie);
        die('nothing');
        return undef;
    }
```

racke:
Fix the defects in the current check:

- Allows bogus session id
- Allows session id which belongs to different user

May fix #1041.

### DIFF
--- a/src/lib/Sympa/WWW/SOAP.pm
+++ b/src/lib/Sympa/WWW/SOAP.pm
@@ -321,19 +321,16 @@ sub authenticateAndRun {
     ## session_table instead
     my $session =
         Sympa::WWW::Session->new($ENV{'SYMPA_ROBOT'}, {cookie => $cookie});
-    if (defined $session) {
-        $email      = $session->{'email'};
-        $session_id = $session->{'id_session'};
-    }
-    unless ($email or $email eq 'unknown') {
-        $log->syslog('err', 'Failed to authenticate user with session ID %s',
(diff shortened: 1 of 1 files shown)


ikedas:
Let's try it out with the next beta.

Issue #596, opened 2019-04-12
ldap ssl connexion no error message


goudal-emmk:
Version
-------

Sympa 6.2.41b.2

Installation method
-------------------

from source

Expected behavior
-----------------

I had some errors in the ldaps configuration (fbad ssl version, and some problem with certificate chain. 
The only error message is that the connexion failed.

Actual behavior
---------------

Errors are generated either by Net::LDAP module or IO::Socket::SSL but none of these errors are treated by sympa who just put a generic message error "connexion failed" and not the precise message from the preceding modules.
What I would expect is to have these message repported by sympa

Additional information
----------------------

goudal-emmk:
Question ? 
It's just a bug concerning error messages not beeing caught and displayed !

racke:
I concur about this issue being a bug. The actually LDAP error should bubble up to the Sympa logs.

ikedas:
@goudal-emmk.  Could you please show us detailed information?  For example logs when error occurred (we at least have to know when does it occur).

racke:
#1045 might be helpful here.

PR #1045, opened 2020-11-29
Add actual error to the log if connection to LDAP server fails.

racke:
This is really helpful to figure out what went wrong with the connection to the LDAP server.


ikedas:
This may fix #596

racke:
I think that fix is a very good candidate for the next beta.

ikedas:
I agree. #596 is assigned the milestone 6.2.60.

Issue #1028, opened 2020-11-03
MySQL: Upgrading fails due to stricter SQL mode


khaoula-abaidi:
Version
-------
From 6.1.24 TO Sympa 6.2.40
Migration  method
-------------------
Transfer .incl files , expl lists , topics , scenari 
Stop services sympa , apache2 ,postfix
/usr/lib/sympa/bin/sympa.pl --upgrade 
Restart services sympa, apache2, postfix

Expected behavior
-----------------
Use of uninitialized value $effective_size in numeric ge (>=) at /usr/share/sympa/lib/Sympa/DatabaseManager.pm line 712.
Use of uninitialized value $effective_size in numeric ge (>=) at /usr/share/sympa/lib/Sympa/DatabaseManager.pm line 712.
Database sympa defined in sympa.conf has not the right structure or is unreachable. verify db_xxx parameters in sympa.conf

Actual behavior
---------------
Erreur interne du serveur

Sympa a été confronté à une erreur interne

Veuillez contacter le listmaster

Erreurs : Can't use an undefined value as an ARRAY reference at /usr/share/sympa/lib/Sympa/List.pm line 7055.
Pile d'appel

DIED: Can't use an undefined value as an ARRAY reference at /usr/share/sympa/lib/Sympa/List.pm line 7055.
 at /usr/share/sympa/lib/Sympa/List.pm line 7055.
	Sympa::List::sync_include_admin(Sympa::List <listexample@xxxx.yyyy.fr>) called at /usr/share/sympa/lib/Sympa/List.pm line 418
	Sympa::List::new('Sympa::List', 'listexample2', 'xxxx.yyyy.fr', HASH(0x55e0ab115f40)) called at /usr/share/sympa/lib/Sympa/List.pm line 7855
	Sympa::List::get_lists('xxxx.yyyy.fr') called at /usr/lib/cgi-bin/sympa/wwsympa.fcgi line 9559
	main::do_get_latest_lists() called at /usr/lib/cgi-bin/sympa/wwsympa.fcgi line 1566

Additional information
----------------------

ikedas:
Hi @khaoula-abaidi ,

> Transfer .incl files , expl lists , topics , scenari

I think you should also transfer the database.

khaoula-abaidi:
> Hi @khaoula-abaidi ,
> 
> > Transfer .incl files , expl lists , topics , scenari
> 
> I think you should also transfer the database.

I've transfered the database also

racke:
[...]

PR #1053, opened 2020-12-08
Update List.pm

faust64:
Upgrade process may fail inserting records in the `list_table` table, due to the `searchkey_list` value overflowing that column size (`varchar(255)`).
Truncates that field, to ensure upgrade doesn't skip anything.

Wild guess / maybe there's other places we could do this.
Found out a mention of that issue over there: https://listes.renater.fr/sympa/arc/sympa-fr/2019-10/msg00017.html
As of patching that file on my server, I can re-run the upgrade (from 5.4.7 to 6.2.40),  without any mention of such failures.

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -6037,7 +6037,7 @@ sub _update_list_db {
 
     my $name = $self->{'name'};
     my $searchkey =
-        Sympa::Tools::Text::foldcase($self->{'admin'}{'subject'} || '');
+        substr( Sympa::Tools::Text::foldcase($self->{'admin'}{'subject'} || ''), 0, 255);
     my $status = $self->{'admin'}{'status'};
     my $robot  = $self->{'domain'};
 


ikedas:
Hi @faust64 , could you please tell us from what version to what version you tried upgrading?

faust64:
> from 5.4.7 to 6.2.40

ikedas:
Sorry, I overlooked your first comment.
I guess this bug was found out due to upgrade of MySQL server (see also #1028).  This bug may also affect to the other DBMSs.  Thank you for reporting and fixing bug!

BTW I have one thing to point out about your fix.  I'll write it in a few days.

ikedas:
BTW I have one thing to point out about your fix.

As the result of ``Sympa::Tools::Text::foldcase()`` can be a multibyte string (for example, the word "humanité" consists of 9 bytes, though it consists of 8 letters), `substr()` might break character boundary.  With MySQL, partial byte sequence is simply ignored, but the other DBMS does not always do such.

It turned out not to be easy to solve this problem.  I'll commit your PR after the next stable and then add fix.

ikedas:
Additional fix has been merged.  @faust64 , thank you for reporting and fixing the bug!

PR #1061, opened 2020-12-21
Skip execution of aliases_program in newaliases script if value is "none"

racke:
This fixes an annoying error message. The problem is more or less cosmetic as the aliases are already probably updated.

```
~$ sympa --close_list test
err main::#144 Invalid aliases_program "none"
close_list [notice] List has been closed

```
Sympa version:  6.2.59b.2

Relevant configuration:
```
sendmail_aliases /etc/mail/sympa/aliases
alias_manager Template
aliases_program none
aliases_wrapper off
```

### DIFF
--- a/src/lib/Sympa/Config/Schema.pm
+++ b/src/lib/Sympa/Config/Schema.pm
@@ -479,11 +479,12 @@ our %pinfo = (
         context    => [qw(domain site)],
         order      => 4.04,
         group      => 'mta',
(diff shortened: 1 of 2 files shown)


ikedas:
Currently there is no option "none" for `aliases_program` parameter.  So this PR may be treated as enhancement rather than bug.

racke:
Yes, as Exim doesn't need an `aliases_program`. It reads the text files.

PR #1062, opened 2020-12-22
Adjust compose_mail template to the obsolescence of the merge_feature configuration parameter

racke:
Otherwise the legend describing the available TT2 parameter doesn't show up. This needs to be mentioned in the upgrade documentation as well.

### DIFF
--- a/default/web_tt2/compose_mail.tt2
+++ b/default/web_tt2/compose_mail.tt2
@@ -53,7 +53,7 @@
             <textarea name="body" id="body" cols="80" rows="25">[% body %]</textarea>
             <br />
 
-            [% IF listconf.personalization_feature == 'on' %]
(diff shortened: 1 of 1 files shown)


default/web_tt2/compose_mail.tt2
```diff
@@ -53,7 +53,7 @@
             <textarea name="body" id="body" cols="80" rows="25">[% body %]</textarea>
             <br />
 
-            [% IF listconf.merge_feature == 'on' %]
+            [% IF listconf.personalization_feature == 'on' %]
```
ikedas, line 56: I suppose it'd be better to be
``` code
            [% IF listconf.personalization_feature == 'on' && listconf.personalization.web_apply_on == 'all' %]
```

racke:
Correct, I fixed that. Thanks for your help.

ikedas:
This is an additional fix for #1037

PR #1068, opened 2020-12-28
Fix regression in personalization feature (#1066)

### LINKED ISSUE #1066: Regression in personalization feature
Only the **body** is processed with _web_apply_on_ set to `all`, but header and footer are skipped. So it will be not possible to restore the current behaviour with release 6.2.60.

racke:
This should fix #1066. Tested with web and email using `footer_type append`.
It is likely that if works for the other possible scenarios as well.


ikedas (approved):
Exactly.  I approve your fix. Thanks!

PR #1069, opened 2020-12-29
Fix regression in personalization feature (#1066): Case 2bis

ikedas:
This may fix (part of) #1066 .


racke (commented):
This indeed fixes the problem. But we are using this block in multiple places which violate the DRY principle. I suggest that we use instead a new method `shelve_personalization` in Sympa::List which takes the type as parameter (web or mail), shelves when appropriate and returns a true value if the message is going to be personalized.

racke:
It would be reasonable though to merge this fix and open a new ticket for the DRY fixes with the next milestone. I would volunteer to do the fixes.

PR #1087, opened 2021-01-19
Remove undocumented backtick syntax in sympa.conf

### LINKED ISSUE #1086: Remove undocumented backtick syntax in sympa.conf
Backtick syntax (`` `...` ``) in `sympa.conf` should be removed, because it allows to execute arbitrary code.

Note: that syntax has not been documented yet.

ikedas:
This may fix #1086

### DIFF
--- a/src/lib/Conf.pm
+++ b/src/lib/Conf.pm
@@ -1728,6 +1728,11 @@ sub _load_config_file_to_hash {
             my ($keyword, $value) = ($1, $2);
             $value =~ s/\s*$//;
 
+            # Deprecated syntax: `command`
+            if ($value =~ /^\`(.*)\`$/) {
+                die sprintf "%s: Backtick (`...`) in sympa.conf is no longer allowed. Check and modify configuration.\n", $value;
+            }
+
             $keyword =
                 $Sympa::Config::Schema::obsolete_robot_params{$keyword}
                 // $keyword;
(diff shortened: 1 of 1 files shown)


racke:
It is very good to remove this "backdoor". But why keep it as comment?

ikedas:
No particular reason. Conf.pm will be removed by radical refactoring anyway, so I thought I'd remind ourselves what _not_ to do in the new code.

racke:
Well, if you ask me we should remove it right away. I'm sure we don't produce code like that anymore.

ikedas:
Ok, I'll do it.

dverdin:
Indeed, it was definitely a trace from the past. If somebody enabled the conf edition on the web interface, it was a way to get/modify data from the whole server; Particularly concerning in the case of virtual domains hosting.
Would it be a good idea to issue a security notice when releasing this change?

racke:
It is still used in the Debian packages, but I will take care of that. 

@ikedas Are there any other known security problems left?

dverdin:
Would it be wrong to make a new tag only for this issue?

ikedas:
> It is still used in the Debian packages, but I will take care of that.
> 
> @ikedas Are there any other known security problems left?

At least problems related to parsing configuration file, I don't know the other problems.

ikedas:
Updated.  Now the process of Sympa terminates if backtick is contained in sympa.conf so that administrators have to update configuration manually.

ikedas:
If no objection, I'll merge this.

Issue #1091, opened 2021-01-21
Obsolete `cookie` parameter


ikedas:
Expected Behavior
-----------------
`cookie` parameter will be obsoleted.

Possible Solution
-----------------

  - [x] Modify codes that are using cookie parameter appropriately.
  - [x] Update documentation.

Context
-------

As far as I investigated with current stable (6.2.60), `cookie` parameter is used _at least_ in these things:

  - bin/upgrade_send_spool.pl#sympa_checksum()
    to migrate older bulk spool
    This is needed for upgrading from versions prior to 6.2.

  - bin/upgrade_sympa_password.pl#_decrypt_rc4_password()
    to migrate older user password
    This is needed for upgrading from versions prior to 6.2.26 by which RC4 encryption was deprecated.

  - Sympa::Tools::Password#tmp_passwd()
    to create temporary password for new users;
    This process should be removed after careful review.

  - Sympa::Archive#_get_tag()
    to generate randomized tags for `$tag$` notation in `mhonarc_ressources.tt2`
    This process should be removed by adopting another randomization.

  - Sympa::List.pm#get_cookie()
    to get List's (not the default in sympa.conf) `cookie` parameter
    This function is no longer used.

racke:
Thanks a lot for the thorough investigation, @ikedas.

For the first item ... I think it is about time to retire direct upgrades from 6.1 to the latest 6.2 release.

ikedas:
It's nice idea to give up direct upgrade from 6.1. (You might mean of Debian packages.)

I think `sympa.pl --upgrade` would be better to stop processing when `sympa.conf` contains backtick.  By this, the sites not using this syntax will be upgraded safely.

racke:
> It's nice idea to give up direct upgrade from 6.1. (You might mean of Debian packages.)
> 

This suggestion was for Sympa itself only. Debian supports updates to the next release only.

> I think `sympa.pl --upgrade` would be better to stop processing when `sympa.conf` contains backtick. By this, the sites not using this syntax will be upgraded safely.

Yes, that makes sense to me.

dverdin:
> For the first item ... I think it is about time to retire direct upgrades from 6.1 to the latest 6.2 release.

[...]

PR #1092, opened 2021-01-21
Remove obsolete script upgrade_send_spool

racke:
Direct update of 6.1 will no longer be supported, so this script isn't needed anymore. See #1091.
Not tested yet.


dverdin:
Hi,
That simplifies certainly the code but how will people upgrade from Sympa prior to 6.2?

racke:
They need to use an intermediary version. I don't see a point in upgrading from a version that is more than five years old.

dverdin:
Sympa handles upgrade from the oldest version possible, i you look at Upgrade.pm. Removing upgrade_send_spool.pl will break this functionality.
I don't understand how to use an intermediate version. They will still need to upgrade the send spool.
Do you remove this script only because of the cookie parameter? Or is it hard to maintain as a packager?

racke:
I really would like to move on with Sympa and not keep archaic code forever. Users upgrading from 6.1 can go first to any version in 6.2. It is ridiculous to carry around that code for a few users which may have affected and sitting on a stale and insecure Sympa version.

dverdin:
I certainly understand the argument about archaic code.
However, if they have old versions, it means they are long-term users.
I don't think it is fair to leave them without any solution to move to the newest version.
The script should at least move to contributions, with notes about how to use it.

racke:
The solution is to upgrade to 6.2.60 first and then to 6.2.xx. How can that be a problem?

dverdin:
OK, got it. Correctly documented, people could still upgrade, even from very old versions.
I still think that the question of version upgrade deserves an issue of its own.

ikedas:
> The solution is to upgrade to 6.2.60 first and then to 6.2.xx. How can that be a problem?

I once thought of putting such "check point", too.  However, I rethought:
  - Codes for measure on incompatibility are limited in `Sympa::Upgrade` and support tools under `src/bin`.
  - `Sympa::Upgrade` will be updated incrementally.  This means that gone versions won't affect to changes in the future.
  - Why 6.2.x lasts for long time is that quality of code is too poor to change focus of development to the new major release.  There are no reason to add a discontinuity at midst of 6.2.x.

In short, I think we  should dedicate our time to remove harmful things, keeping harmless things untouch.

racke:
These things are not harmless, they are wasting time when navigating or search the source code.

racke:
And I really can't fathom a new major release at this point.

ikedas:
> These things are not harmless, they are wasting time when navigating or search the source code.


> And I really can't fathom a new major release at this point.

Honestly, I also can't imagine new major release as an extension of the current work  (At least some components including wwsympa would be faster to rewrite, if we were willing).

Besides, how about submitting new issue to consider what should be removed from recent 6.2? There may also be the others than codes for upgrade.

racke:
Apparently @ikedas and @dverdin want to keep this script against reasonable arguments.

PR #1108, opened 2021-02-01
`web_archive_spam_protection gecos` didn't work as expected

### LINKED ISSUE #1107: Protect web archive against spam harvesters: "only show gecos" doesn't seem to work
Version
-------
6.2.60

Installation method
-------------------
rpm

Expected behavior
-----------------

A message with:

```From: David Carter <dpc22@cam.ac.uk>```

should be displayed as:

```From: David Carter```

in the list archives  when the web_archive_spam_protection list setting is "gecos" (only show Gecos)

Actual behavior
---------------

The full email address is displayed. 

```From: David Carter <dpc22@cam.ac.uk>```

If I pick: "at" (replace @ characters) instead, I see:
 
```From: David Carter <dpc22 AT cam.ac.uk>```

which is what I would expect

Additional information
----------------------

None

### PR DESCRIPTION
**Pull request by ikedas**
`web_archive_spam_protection gecos` didn't work because this option was omitted.  This PR may fix #1107 .

This PR includes the other fixes:
  - Previously email addresses in HTML tags were not removed.
  - The option name "gecos" is not so generally used. "concealed" looks more common.

### DIFF
--- a/src/lib/Sympa/HTMLDecorator.pm
+++ b/src/lib/Sympa/HTMLDecorator.pm
@@ -249,6 +249,7 @@ sub decorate_email_concealed {
         } elsif ($item->{event} eq 'start'
             and $item->{attr}
             and 0 == index(lc($item->{attr}->{href} // ''), 'mailto:')) {
+            # Empties mailto URL in link target
             my $text = $item->{text};
             $text =~ s{(?<=\bhref=)[^\s>]+}{"mailto:"}gi;
             $decorated .= $text;
--- a/src/lib/Sympa/HTMLDecorator.pm
+++ b/src/lib/Sympa/HTMLDecorator.pm
@@ -182,13 +182,12 @@ sub decorate {
 
     return $html unless defined $html and length $html;
 
-    if ($options{email}) {
-        $self->{_shdEmailFunc} =
-              $options{email} eq 'at'         ? \&decorate_email_at
-            : $options{email} eq 'gecos'      ? \&decorate_email_gecos
-            : $options{email} eq 'javascript' ? \&decorate_email_js
-            :                                   undef;
-    }
+    $self->{_shdEmailFunc} = {
+        at         => \&decorate_email_at,
+        concealed  => \&decorate_email_concealed,
+        gecos      => \&decorate_email_concealed,    # compat.<=6.2.61b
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/lib/Sympa/HTMLDecorator.pm [hunk 1]
```diff
@@ -247,13 +246,17 @@ sub decorate_email_gecos {
             } else {
                 $decorated .= $item->{text};
             }
+        } elsif ($item->{event} eq 'start'
+            and $item->{attr}
+            and 0 == index(lc($item->{attr}->{href} // ''), 'mailto:')) {
+            my $text = $item->{text};
+            $text =~ s{(?<=\bhref=)[^\s>]+}{"mailto:"}gi;
```
ldidry, line 253: Can you explain me the regex? I don’t get it :thinking:

ldidry (commented):
Looks good to me, but I would like to understand the regex in HTMLDecorator.pm.

src/lib/Sympa/HTMLDecorator.pm
ikedas: `(?<=`...`)` is looking-behind assertion.  Matching will succeed if the pattern that assertion includes precedes (See [tutorial](https://perldoc.perl.org/perlretut#Looking-ahead-and-looking-behind) for details). 

`(?<=\bhref=)[^\s>]+` may be read as "a string matching with `[^\s>]+` preceded by a string matching with `\bhref=`".  Thus, in a string
> `<a href="mailto:me@example.org">`

only `"mailto:me@example.org"` matches: Assertion is not included in the range of match.

Therefore, this line replaces the value of `href` attribute in the tag with a empty mailto URL "`mailto:`".

ldidry: Ok, thanks. I never used looking-behind assertion, so didn’t knew the syntax.

It could be useful for future people like me to add a comment like `# Emptying mailto link target`, don’t you think? That’s just a suggestion, not mandatory.

ikedas: > It could be useful for future people like me to add a comment like `# Emptying mailto link target`, don’t you think? That’s just a suggestion, not mandatory.

Makes sense.  I’ll do it.  Thanks for advice!

PR #1112, opened 2021-02-04
bulk.pl daemon crashes if it failed to load private key (#1110)

### LINKED ISSUE #1110: Moderation mail that could not be sent blocked all outgoing mail
We recently had the problem that a moderation email could not be sent to the owners of a list for which there is a certificate and through which encrypted emails can be sent.
The error was that the correct value for the key_passwd parameter was not set in the Sympa config and the error message "failed to load the private key" appeared in the Sympa log.
This was my error and caused the email to the list moderators to get stuck in the queue /var/spool/sympa/bulk/msg/.

The problem with this situation, however, was that during the time Sympa kept trying to send this email, the emails to subscribers of all the other mailing lists were not being sent.
So the mailing list server was accepting and processing emails for the lists, but the emails to the subscribers of the list were sent only after we detected and fixed the above problem.

Do you have any idea why this one email blocked the sending of emails to the subscribers of all other lists?
It looks like this one email blocked the entire bulk spool.
Of course, we would like to avoid that in a similar case in the future the outgoing mail dispatch could be blocked again.

We use Sympa version 6.2.16 and Sympa is installed as Debian Package.

Kind regards,
Sabine

ikedas:
- bulk.pl daemon crashes if it failed to load private key to sign a message.
  - There is also a case with public key (with 6.2.16: recent version has been fixed).

This may fix #1110 (the first commit is done in relative to Sympa 6.2.16).


racke (approved):
Looks good to me.

Issue #1033, opened 2020-11-16
Chrome incorrectly auto-completes "include_file" using email address of listmaster


dpc22:
Version
-------
6.2.58

Installation method
-------------------

RHEL 7 RPM based on:

https://dl.fedoraproject.org/pub/epel/7/SRPMS/Packages/s/sympa-6.2.58-1.el7.src.rpm

Expected behaviour
-----------------

Should be able to click "Apply modifications" button at bottom of 

  Edit List Congigurations
    -> Data sources setup

without automatically adding an "include_file" which corresponds to email address of listmaster. 

Unsurprisingly that doesn't work, and an error is reported whenever you try to synchronise the data source.

Actual behaviour
---------------

include_file set to email address used to log in as listmaster unless I manually clear the value of  "include_file" before clicking "Apply modifications".

Additional information
----------------------

The first time that I  log into Sympa as the listmaster, most Web browsers (at least Chrome/Firefox/Safari) offer to save the username and password combination.

If I then navigate to:

 Edit List Congigurations
    -> Data sources setup

the saved username and password values auto-completes various fields.
 
In Chrome:

File inclusion
  (include_file)(default)
￼Remote file inclusion (include_remote_file)(default)
  remote password (passwd)

In Firefox and Safari:

Remote file inclusion (include_remote_file)(default)
  remote user (user)
￼  remote password (passwd)

Which is still wrong, but at least has no effect when you click "Apply modifications", unless some information is provided for the other two attributes in the same section, namely:

 short name for this source (name)
￼ data location URL (url)

It isn't clear to me whether the problem is with Chrome or Sympa, but it would be useful if we could stop Web browsers from trying to auto-fill these fields in the first place. It looks like we just need to set a HTML attribute "autocomplete=off" :

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

ikedas:
FYI:
- [How to turn off form autocompletion - MDN](https://developer.mozilla.org/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion)
- [Disabling Chrome Autofill - Stack Overflow](https://stackoverflow.com/questions/15738259/disabling-chrome-autofill)

dpc22:
Thanks.

We are probably going to have quite a lot of listmasters (5 to 10?) compared to most Sympa installers so that our service desk front line staff can create/approve new mailing lists. 

[...]

PR #1114, opened 2021-02-05
Add autocomplete="off" to all forms unless explicitly enabled.

ikedas:
### Known bug
  - However, this cannot prevent heuristics by Google Chrome: See #1033 .

### DIFF
--- a/default/web_tt2/login.tt2
+++ b/default/web_tt2/login.tt2
@@ -63,7 +63,7 @@
     [% END ~%]
 
     [% IF use_passwd ~%]
-        <form action="[% path_cgi %]" method="post">
+        <form action="[% path_cgi %]" method="post" autocomplete="on">
             <fieldset>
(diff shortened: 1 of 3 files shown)


racke:
Can you please explain why autocomplete needs to be off on all forms?

ikedas:
> Can you please explain why autocomplete needs to be off on all forms?

Not all.

W3C/WAI WCAG21 [recommends](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html) that "each input field collecting information about the user" implements technology to support inputs, such as auto-completion.  That's why this PR enables auto-completion on `login` and `renewpasswd` forms.

However, auto-completion futures currently common browsers implement are not suitable for the other forms: Those features support input of information about the user of the browser itself, while those form (e.g. subscribers management) may used to input the information about the other users.  Therefore, enabling such features can be a securirty risk.

Issue #876, opened 2020-02-18
Death of the wwsympa process bug


bruncsak:
Version
-------
sympa-6.2.52-2.el7.x86_64

Installation method
-------------------
via yum on CentOS 7 from the epel repository

I got regularly the following error message:
````
wwsympa[1234]: err main::#1544 > main::do_help#3868 > Sympa::get_url#756 DIED: Can't locate object method "host_port" via package "URI::_generic" at /usr/share/sympa/lib/Sympa.pm line 756.
````
I have in the apache error log the following in addition too:
````
[<date>] [fcgid:warn] [pid num1] [client 1.2.3.4:5678] mod_fcgid: stderr: Use of uninitialized value in string eq at /usr/share/sympa/lib/Sympa.pm line 753.
[<date>] [fcgid:warn] [pid num2] mod_fcgid: cleanup zombie process 1234
````

ikedas:
Hi @bruncsak ,
- Could you please apply [this patch](https://github.com/sympa-community/sympa/pull/878.patch) and check if the problem will be solved?
- What is the value of `wwsympa_url` parameter in your `sympa.conf` (or `robot.conf`)?

bruncsak:
I applied the patch, we will see is the problem still there, or not any more, thanks.

Concerning wwsympa_url:
There is no such entry in /etc/sympa/sympa.conf, I am using exclusively robot.
In the robot.conf the format of the wwsympa_url entry is simmilar to:
`wwsympa_url http://www.example.com/aa/bb`
The `www.example.com` hostname in the URL is not the hostname where the sympa itself is running, there is a reverse-proxy in front of the sympa host.

I am using non-standard configuration relative to the rpm package from the EPEL repo, due to the double components of the wwsympa_url local path. I use mod_fcgid instead of mod_proxy_fcgi. For additional detail please visit:
https://bugzilla.redhat.com/show_bug.cgi?id=1770783

bruncsak:
There is one more information. I just realized that the configuration is incorrect on our side. The reverse proxy actually unloads the SSL tunnel, so the correct value for wwsympa_url should have been:
`wwsympa_url https://www.example.com/aa/bb`
instead of `http` schema.
Is it correct to write:
`wwsympa_url //www.example.com/aa/bb`, with other term not specifying the URL schema?

ikedas:
@bruncsak , please submit new report for another issue. As I submitted #879, please post additional comment if any.

ikedas:
[...]

PR #1115, opened 2021-02-07
Make it clear that wwsympa_url needs to include the protocol

racke:
This prevents confusion seen at:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=929157#15
https://github.com/sympa-community/sympa/issues/876#issuecomment-590156276

Even better to have a format that checks whether it is a valid URL.
E.g. by feeding it to the URI module and verify that all vital parts
are present.

### DIFF
--- a/src/lib/Sympa/Config/Schema.pm
+++ b/src/lib/Sympa/Config/Schema.pm
@@ -4810,7 +4810,7 @@ our %pinfo = (
         group           => 'www_basic',
         sample          => 'https://web.example.org/sympa',
         gettext_id      => 'URL prefix of web interface',
-        gettext_comment => 'This is used to construct URLs of web interface.',
+        gettext_comment => 'This is used to construct URLs of web interface. The protocol (either https:// or http://) is required.',
     },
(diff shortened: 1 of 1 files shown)


ikedas:
> This prevents confusion seen at:
> 
> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=929157#15
> [#876 (comment)](https://github.com/sympa-community/sympa/issues/876#issuecomment-590156276)

I have read this bugreport and at the time felt good idea to omit schema in URL (it's a legitimate notation of relative URI reference: See [RFC3986, 4.2](https://tools.ietf.org/html/rfc3986#section-4.2)).  But now I know it's impossible.

This omission is useful in HTML source, because protocol of URLs may be taken from base URL the browser is accessing.  However, it cannot apply to the parameter(s) of Sympa, because the server has to give the protocol used by itself.

> Even better to have a format that checks whether it is a valid URL.
> E.g. by feeding it to the URI module and verify that all vital parts
> are present.

I agree.

PR #1133, opened 2021-03-22
Skip approval for owner subscription and unsubscription

racke:
There is no point to force the list owner to approve its own subscription and unsubscription to a list.


racke:
This problem occurs when the list owner uses "Subscribe" or  "Unsubscribe" in the list menu.

Issue #978, opened 2020-08-03
Listmaster notification sent for sync_include_failed


Vincent3187:
Hello,
Some of our lists configured with datasources (remote file or LDAP) fail to synchronize members but no notification corresponding is sent to listmaster. The corresponding error is displayed in the logfile, for example :
Unable to fetch data source Sympa::DataSource::RemoteFile ....Can't connect to…
Can we have accidentally disabled this type of notification ? We used to receive them few weeks ago.

Sympa 6.2.54

Thanks for your help

ikedas:
Hi @Vincent3187 ,

Could you please show us the full log lines indicating the errors?

Vincent3187:
Hello Ikedas,
These are 3 examples of error :

Example with remote file :
Aug  3 01:57:12 tcgpsympa2 task_manager[91000]: err main::#148 > Sympa::Spindle::spin#83 > Sympa::Spindle::ProcessTask::_twist#74 > Sympa::Spindle::ProcessTask::_execute#140 > Sympa::Spindle::ProcessTask::_cmd_process#174 > Sympa::Spindle::ProcessTask::do_sync_include#1349 > Sympa::List::sync_include#4798 > Sympa::Spindle::spin#95 > Sympa::Request::Handler::include::_twist#181 > Sympa::Request::Handler::include::_update_users#289 > Sympa::DataSource::open#134 > Sympa::DataSource::RemoteFile::_open#101 Unable to fetch data source Sympa::DataSource::RemoteFile <context=XXXXXXX@services.cnrs.fr;id=bcb241c6;role=member;name=XXXX>: Can't connect to XXXX:80 (Connexion refusée)

Example with LDAP :
[...]

PR #1140, opened 2021-04-01
Add list visibility scenario for logged in users.

racke:
This scenario was created for one of my customers and is IMHO a sensible addition, especially in a corporate environment.


Playaset (commented):
#1904 #1904 #978 #1898

PR #1183, opened 2021-06-07
Make distribution_ttl and inclusion_notification_feature site-wide

Jylhis:
Also add site default for distribution_ttl

### DIFF
--- a/src/lib/Sympa/Config/Schema.pm
+++ b/src/lib/Sympa/Config/Schema.pm
@@ -3519,7 +3519,7 @@ our %pinfo = (
     },
 
     distribution_ttl => {
-        context => [qw(list)],      #FIXME: No site-wide default
+        context => [qw(list site)],
         order   => 60.13,
         group   => 'data_source',
         gettext_id => "Inclusions timeout for message distribution",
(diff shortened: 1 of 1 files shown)


src/lib/Sympa/Config/Schema.pm
```diff
@@ -3519,14 +3519,15 @@ our %pinfo = (
...
-        context => [qw(list)],      #FIXME: No site-wide default
+        context => [qw(list site)],
         order   => 60.13,
         group   => 'data_source',
         gettext_id => "Inclusions timeout for message distribution",
         gettext_comment =>
             "This parameter defines the delay since the last synchronization after which the user's list will be updated before performing either of following actions:\n* Reviewing list members\n* Message distribution",
         gettext_unit => 'seconds',
         format       => '\d+',
+        default      => '300',
```
ikedas, line 3530: The only place this parameter is used is:
https://github.com/sympa-community/sympa/blob/39e3006ad6c948dba9e8322f9d5a5823ff2d727b/src/lib/Sympa/Spindle/ToList.pm#L194-L198
Thus I think the default value of `distribution_ttl` should _not_ be defined so that the value of `ttl` may be used.

Jylhis: I now pushed changes with the default value definition removed.

ikedas: Thanks!

ikedas (approved):
Thank you for improvement!

PR #1184, opened 2021-06-07
Data source synchronization performance improvement

Jylhis:
**[Note by admin]** Please visit #1186 for discussion.

----
I've been looking into improving the performance of synchronizing members from data sources. I moved a couple of queries outside of loops and enabled transactions for the insert/update/remove loops.

This PR currently work in progress. It is lacking some error handling and it fails with SQLite, because of the current SQLite locking workaround.

I would like comments if this type of changes would be acceptable and suggestions for improvements!

# Benchmarks
### add 20 000, remove 20 000 update 20 000 members
| |Before | After |
|-|-------|-------|
|Time (sec)| 263 | 67.2 |
| Calls to do_prepared_query |240 032 | 120 032  |

### add 60 000, update 40 000 members
| |Before | After |
|-|-------|-------|
|Time (sec)| 569| 141|
| Calls to do_prepared_query |580 032 | 260 032 |

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -1547,7 +1547,7 @@ sub delete_list_member {
     my $total = 0;
 
     my $sdm = Sympa::DatabaseManager->instance;
-    $sdm->__dbh->begin_work;
+    $sdm->begin;
 
     foreach my $who (@u) {
         $who = Sympa::Tools::Text::canonic_email($who);
@@ -1602,13 +1602,12 @@ sub delete_list_member {
     $self->_cache_publish_expiry('member');
     delete_list_member_picture($self, shift(@u));
 
-    unless ($sdm->__dbh->{AutoCommit}) {
-        my $rc = $sdm->__dbh->commit;
+        my $rc = $sdm->commit;
         unless ($rc) {
             $log->syslog('err', 'Error at delete member commit: %s', $sdm->errstr);
-            $sdm->__dbh->rollback;
+            $sdm->rollback;
         }
-    }
+
     return (-1 * $total);
 
 }
@@ -3226,7 +3225,7 @@ sub add_list_member {
 
     my $sdm = Sympa::DatabaseManager->instance;
 
-    $sdm->__dbh->begin_work;
+    $sdm->begin;
 
     foreach my $new_user (@new_users) {
         my $who = Sympa::Tools::Text::canonic_email($new_user->{'email'});
@@ -3373,13 +3372,12 @@ sub add_list_member {
         $self->{'add_outcome'}{'remaining_member_to_add'}--;
         $current_list_members_count++;
     }
-    unless ($sdm->__dbh->{AutoCommit}) {
-        my $rc = $sdm->__dbh->commit;
+        my $rc = $sdm->commit;
         unless ($rc) {
             $log->syslog('err', 'Error at add member commit: %s', $sdm->errstr);
-            $sdm->__dbh->rollback;
+            $sdm->rollback;
         }
-    }
+
 
     $self->_cache_publish_expiry('member');
     $self->_create_add_error_string() if ($self->{'add_outcome'}{'errors'});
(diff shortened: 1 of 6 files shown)


ikedas:
Hi @Jylhis ,
I think we'd be better to have a new issue to discuss on your proposal and related PRs.  Could you please [do it](https://github.com/sympa-community/sympa/issues/new/choose)?

Jylhis:
Hi @ikedas ,
I created the issue with some more information on our case.

src/lib/Sympa/List.pm
```diff
@@ -1570,6 +1572,7 @@ sub delete_list_member {
             $log->syslog('err', 'Unable to remove list member %s', $who);
             next;
         }
+        delete_list_member_picture($self, shift(@u));
```
ikedas, line 1575:
``` perl
        delete_list_member_picture($self, $who);
```
may correct.

ikedas: I'd like to submit the change on this line as a separate PR to fix a bug.

ikedas: Done. See #1189

src/lib/Sympa/Request/Handler/include.pm
```diff
@@ -367,85 +367,166 @@ sub _update_users {
...
+                    next;
+                }
+            }
+
+            # 3. Update
+            my %res = __update_user($ds, $email, $start_time, $list, $t, $r, $is_external_ds, $time);
+
+            unless (%res) {
+                $ds->close;
+                $log->syslog('info', '%s: Aborted update', $ds);
```
ikedas, line 455: Maybe this is needed here.
``` perl
            $sdm->__dbh->rollback
```

ikedas:
Conflict between this pr and #1189 was resolved.

ikedas:
Some other comments:
- `add_list_admin` and `delete_list_admin` may also be wrapperd by begin-commit.
- Typos: "`$sdm->errstr`" should be "`$sdm->error`".

Jylhis:
`add_list_admin` is little bit trickier to wrap in begin-commit because if the `replace` option is set then the function tries to do UPDATE and INSERT. There might be a case when both UPDATE and INSERT would be executed successfully.

https://github.com/sympa-community/sympa/blob/978001b8101674a17754767e619d4fda0a1f2688/src/lib/Sympa/List.pm#L3488-L3529

I also noticed that inside the `User` constructor there is call to `get_global_user` function, which does one SELECT. As far as I know this query would not be executed at the right time when transactions are used. This didn't cause any errors in my tests with `add_list_member` before, so I'm not sure if it's necessary or I just didn't happen to hit any cases where problem would be triggered.

https://github.com/sympa-community/sympa/blob/978001b8101674a17754767e619d4fda0a1f2688/src/lib/Sympa/List.pm#L3301-L3306

and the query:

https://github.com/sympa-community/sympa/blob/978001b8101674a17754767e619d4fda0a1f2688/src/lib/Sympa/User.pm#L495-L523

ikedas:
> `add_list_admin` is little bit trickier to wrap in begin-commit because if the `replace` option is set then the function tries to do UPDATE and INSERT. There might be a case when both UPDATE and INSERT would be executed successfully.

Please tell me, what exactly are the cases in which both UPDATE and INSERT would be executed successfully?

> I also noticed that inside the `User` constructor there is call to `get_global_user` function, which does one SELECT. As far as I know this query would not be executed at the right time when transactions are used. This didn't cause any errors in my tests with `add_list_member` before, so I'm not sure if it's necessary or I just didn't happen to hit any cases where problem would be triggered.

Maybe I don't understand the problem, but executing SELECT doesn't seem to cause a problem as long as the same connection is kept alive (`$sdm` is a singleton instance, that is, the same connection can be obtained no matter how many times the `->instance` method of `Sympa::DatabaseManager` is called.).

ikedas:
No need to worry about that.  Updates during a transaction appear to be reflected immediately, however updates after BEGIN may be recovered (cancelled) by ROLLBACK or by breakage of connection without COMMIT.  On the other hand, updates can be invisible to _the other sessions_ (the connection by the other process) until COMMIT is executed.

PR #1218, opened 2021-07-31
Make the characters used for e-mail addresses conform to RFC 5322

### LINKED ISSUE #1217: Sympa does not like "`" chars in email addresses within include file
I noticed error messages like this when sympa tried to sync an include file that had a "`" in the local part of the address.  At first I was surprised that Office365 (location of our mailboxes) allowed it, but then I discovered that it was technically OK with the RFCs.  (I can re-dig up the references I found if necessary.)

```
Jun 25 23:20:14 sympa01 sympa[53065]: err main::#849 > Sympa::Spindle::spin#95 > Sympa::Request::Handler::include::_twist#201 > Sympa::Request::Handler::include::_update_users#372 > Sympa::DataSource::next#149 > Sympa::DataSource::File::_next#75 Skip badly formed line: "first1`.last1@x.org First1` Last1"

```

Version
-------
6.2.60

Installation method
-------------------
Source (that locally build into our own rpm)

Expected behavior
-----------------
I think so long as the "`" is in the local part, it should be considered ok, despite it being a bit odd.

Actual behavior
---------------
The sync of the include file never completes successfully.

Additional information
----------------------

ikedas:
Make the characters used for e-mail addresses conform to RFC 5322 ~~, and allow any characters beyond US-ASCII in the local part.  The latter may partially support internationalized e-mail~~.

This may ~~also~~ fix #1217.

EDIT: Support for internationalized e-mail addresses was whithdrawn for now.


rahmatofficialbabu99-sys:
😡

rahmatofficialbabu99-sys:
🙏🏻🙏🏻

Issue #1090, opened 2021-01-20
Slow updates to lists (edit_list_conf cache not working)


dpc22:
Version
-------

6.2.60

Installation method
-------------------

RHEL 7 rpm

Expected behaviour
--------------------

Updates to list configuration via wwsympa  "Admin -> Edit List Config"  (edit_list_request) should take less than a second to complete.

Actual behaviour
---------------

Updates take 5 seconds to complete. (Dedicated test machine. Zero load).

Additional information
----------------------

If I enable logging, almost all of the time is spent in:

/usr/share/sympa/lib/Sympa/List/Config.pm
```
sub get_schema {
...
        foreach my $pname (CORE::keys %{$pinfo || {}}) {
            $self->_get_schema_apply_privilege($pinfo->{$pname}, [$pname],
                $user, undef);
        }
```
Namely _get_schema_apply_privilege:
```
    my $priv = $list->may_edit(join('.', @{$pnames || []}), $user);
```
 which calls /usr/share/sympa/lib/Sympa/List.pm ::
```
sub may_edit {
...
   my $edit_list_conf =
           $self->_cache_get('edit_list_conf')
        || $self->_cache_put('edit_list_conf', $self->_load_edit_list_conf)
        || {};
```

Testing reveals that the cache is never actually used:

```
sub _cache_get {
    my $self = shift;
    my $type = shift;

    my $lasttime = $self->{_mtime}{$type};
    my $mtime;
    if ($type eq 'total' or $type eq 'is_list_member') {
        $mtime = $self->_cache_read_expiry('member');
    } else {
        $mtime = $self->_cache_read_expiry($type);
    }
    $self->{_mtime}{$type} = $mtime;

    return undef unless defined $lasttime and defined $mtime;
    return undef if $lasttime <= $mtime;
    return $self->{_cached}{$type};
}
```

$self->-cache_get() is comparing the mtime on the relevant file with the mtime of the same file on the previous call.
Consequently ```($lasttime <= $mtime)``` in the following is always true:
```
    return undef if $lasttime <= $mtime;
```
unless the file is updated in the middle of an get_schema() call (which should almost never happen, and is the reverse of the desired behaviour for a cache).

I attach a patch which seems closer to the intended behaviour, including a debug statement:

[List.pm.patch.txt](https://github.com/sympa-community/sympa/files/5844498/List.pm.patch.txt)

This is much faster

dpc22:
I attach a different patch:

[...]

PR #1172, opened 2021-05-13
edit_list.conf should not be loaded multiple times to waste memory

ikedas:
edit_list.conf was loaded multiple times and wasted memory.

~This may fix #1170.~

PR #1223, opened 2021-08-18
move pinfo operations in _load_edit_list_conf

olivov:
In Sympa 6.2.62, there was an update made to move the decision for whether or not to reload edit_list.conf from Sympa::List::may_edit -> Sympa::List::new which seems to be in response to reports of slow list configuration saves happening via https://github.com/sympa-community/sympa/issues/1090

Due to this change, the subroutine _load_edit_list_conf is being called much more frequently. 

We ran some profiling data against our Sympa installation and noticed that inside _load_edit_list_conf, there is the pinfo hash which gathers list config data and is an expensive operation for an install containing ~6000 lists.

With that being said, when we upgraded to 6.2.64 from 6.2.60 (we skipped 6.2.62), we noticed our baseline CPU usage increased by about 5x from ~4% to about 25% over a 24hr period. This change was noticed in our qual environment which has a stale copy of our production data but NO customer traffic.

In production, when we upgraded to 6.2.64, we saw an average CPU usage of about ~50% throughout a 24hr period. Which is significantly higher than the CPU usage average we saw in 6.2.60 of ~16%.

The patch provided is very simple and moves the data gathering for the pinfo hash to *after* the decision being made for whether or not the edit_list.conf needs to be reloaded. With this update, our CPU usage over a 24hr period has returned to a nominal state of what we saw in 6.2.60.

We've tested this patch against 6.2.64 with no issues to mention. Please do let us know if this seems workable. Thanks y'all!

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -6081,11 +6081,6 @@ sub _load_edit_list_conf {
 
     my $robot = $self->{'domain'};
 
-    my $pinfo = {
-        %{Sympa::Robot::list_params($self->{'domain'})},
-        %Sympa::ListDef::user_info
-    };
-
     # Load edit_list.conf: Track by file, not domain (file may come from
     # server, robot, family or list context).
     my $last_path_config  = $self->{_path}{edit_list} // '';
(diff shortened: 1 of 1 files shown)


ikedas:
Hi.  Please see also #1172.

olivov:
Excellent! Thanks for pointing that out, Soji. I had seen that PR previously but mistakenly overlooked it. I think the updates in there look to be exactly what we need. I'll test those changes in our environment and report back the findings.

ikedas:
Thanks @olivov !  I'll close this PR.  Please continue with my PR.

PR #1227, opened 2021-08-24
Update fr.po

Pada0:
add an "S" to message


ikedas:
Hi @Pada0 , thank you for proposing correction of translation.

Could you please use [translation site](https://translate.sympa.org/projects/sympa/) to improve translation? Thank you again.

PR #1269, opened 2021-10-29
Build process: Use configured Perl on "make tidyall"

ikedas:
Use Perl specified by --with-perl configure option on "make tidyall".


ldidry (approved):

ikedas:
@ldidry , thanks for approval.  Merge it.

PR #1287, opened 2021-12-05
Adding the workflow for automated testing

ikedas:
Adding GitHub Actions workflow for automated testing.

---
This PR depends on #1288 & #1299 .

### DIFF
--- a/.github/workflows/make-check.yml
+++ b/.github/workflows/make-check.yml
@@ -9,7 +9,7 @@ on:
       - '*'
 
 jobs:
-  build:
+  make-check:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


ikedas:
Okey, it looks working fine.  Let's merge it and see what happens.

racke:
I expect that to fix it for my Debian package workflow as well. This errors out for Perl 5.20. Thanks!

ikedas:
> I expect that to fix it for my Debian package workflow as well. This errors out for Perl 5.20. Thanks!

Yes, I confirmed the `--configure-args` option fixes error with Perl 5.20, if it is added to `cpanm` directly.  I don't know why it is not effective when it is used in workflow...

PR #1289, opened 2021-12-06
fix: create a list as copy does not copy all configuration (template files)

igoltz:
If creating a new list as copy of an existing one then this should also 
1. copy multipart mime footer/header (*.mime files)
2. copy template directories 

The fix (PR) does this by adding both *.mime files to the list of file.
And by fixing the copy logic to also copy template directories.


ikedas:
Hi @igoltz ,
Thank you for fix the bug!  I'll merge your PR with some more fix by me.

PR #1307, opened 2021-12-25
Improvements on anonymized messages

### LINKED ISSUE #1306: SpamAssassin rejects anonymized messages originally posted by Mozilla
Version
-------

Maybe any of 6.2.66 and earlier.

Installation method
-------------------

Maybe any.

Expected behavior
-----------------

Anonymized messages (by `anonymous_sender` parameter enabled) should not be rejected by content filters.

Actual behavior
---------------

SpamAssassin rejects the message with following status:
``` code
X-Spam-Status: No, score=0.313 required=6.31 tests=[AWL=0.643, BAYES_00=-1.9,
 FORGED_MUA_MOZILLA=2.309, HEADER_FROM_DIFFERENT_DOMAINS=0.25, 
 HTML_IMAGE_RATIO_02=0.001, HTML_MESSAGE=0.001, MAILING_LIST_MULTI=-1, 
 NO_RELAYS=-0.001, T_KAM_HTML_FONT_INVALID=0.01] autolearn=no autolearn_force=no
```
The message is rejected mainly due to `FORGED_MUA_MOZILLA` rule. It means that, though the message seems posted with Mozilla Thunderbird, the format of `Message-ID` field looks unfamiliar.

Additional information
----------------------

Originally reported on a thread in `sympa-users`:
https://listes.renater.fr/sympa/arcsearch_id/sympa-users/2021-12/2b094340-6b60-4254-7030-cd1621385f1d%40prompt.nl

ikedas:
* Remove more header fields in anonymized messages, including User-Agent.
* Use more precise number in auto-generated message IDs to prevent collision more.
* Use sufficiently unique message ID above in the anonymized messages.

This may fix #1306 .

### DIFF
--- a/src/lib/Sympa.pm
+++ b/src/lib/Sympa.pm
@@ -46,6 +46,7 @@ use Sympa::Log;
 use Sympa::Regexps;
 use Sympa::Spindle::ProcessTemplate;
 use Sympa::Tools::Text;
+use Sympa::Tools::Time;
 
 my $log = Sympa::Log->instance;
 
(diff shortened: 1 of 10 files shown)


ikedas:
The header fields to be added to the default of `anonymous_header_fields` are taken from:

  - [RFC2076, 3.4](https://datatracker.ietf.org/doc/html/rfc2076#section-3.4) "Sender and recipient indication":

    Fax
    Telefax
    Phone
    Mailer
    Originating-Client
    X-Mailer
    X-Newsreader

  - Additions by [draft-palme-mailext-headers](https://datatracker.ietf.org/doc/html/draft-palme-mailext-headers-08):

    X-MimeOLE
    User-Agent
    X-Face

taggart:
Good idea to strip those others, hopefully that will cover all other potential cases.
One thing I thought of is that since MUAs (and list archives) rely on Message-ID for threading purposes, we should check that threading does the right thing with these improvements. I think the old sequence system sort of worked for replies within a day, but maybe had weird corner cases. I think the new method should work much better.

ikedas:
As far as I can tell, the rewriting of message IDs is done before the messages are archived, so I guess it won't affect threads.

I was planning to release 6.2.68 this week, but if you are willing to take the time to check my changes, the release may be extended by a week or two.

PR #1313, opened 2022-01-06
Remind name of the opened list in the notification

ldidry:
When opening pending list, we have a notification `Aliases have been installed.` but it may be useful to have a reminder of the list we opened (when mass opening the pending lists, for ex.).

### DIFF
--- a/default/mail_tt2/report.tt2
+++ b/default/mail_tt2/report.tt2
@@ -487,7 +487,7 @@
   [%|loc%]List name has been lowercased[%END%]
 
 [%~ ELSIF report_entry == 'auto_aliases' ~%]
-  [%|loc%]Aliases have been installed.[%END%]
+  [%|loc%]Aliases have been installed.[%END%] ([% report_param.listname %])
 
 [%~ ELSIF report_entry == 'user_notified' ~%]
   [%|loc(report_param.notified_user)%]User %1 has been notified[%END%]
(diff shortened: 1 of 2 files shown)


default/mail_tt2/report.tt2
```diff
@@ -487,7 +487,7 @@
   [%|loc%]List name has been lowercased[%END%]
 
 [%~ ELSIF report_entry == 'auto_aliases' ~%]
-  [%|loc%]Aliases have been installed.[%END%]
+  [%|loc%]Aliases have been installed.[%END%] ([% report_param.list_name %])
```
ikedas, line 490: `[% list.name %]` and `[% domain %]` seem to be defined.  cf. [Sympa::Message::Template](https://github.com/sympa-community/sympa/blob/b17e3bb89b904d90704978afe3b9e24bb32f4cf1/src/lib/Sympa/Message/Template.pm#L142-L173)

ldidry: ` [% domain %] ` work when using the web interface, not with `sympa.pl --open_list=list@example.org`, and `[% list.name %]` don’t work at all.

ikedas: Ok, I'll investigate it in the next week (fyi, I found usage of `[% listname %]` but no `[%list_name%]`).

ldidry: Do you prefer `listname` over `list_name`? I have no pb to change it.

ldidry: Changed to `listname`

PR #1314, opened 2022-01-06
🐛 — Protect email addresses in archive search

### LINKED ISSUE #1312: Protected web archive shows email address on search despite of "only show gecos" parameter
Version
-------
6.2.60

Expected behavior
-----------------
Address shouldn't be displayed in search pages.

Actual behavior
---------------
With **Protect web archive against spam harvesters** set on **only show gecos**. Search for a **name** in archive advanced search: email address are displayed.

If you browse into archives (without searching) email addresses are replaced as expected.

ldidry:
Should fix #1312

Warning: when the search matches an email address, it is not obfuscated since the search split the result according to the matching parts. 
Searching for `bar` on `foobar@baz.com` does, for ex:
- `foo`
- `bar`
- `@baz.com`

And none of those parts matches the email regex used to obfuscate email addresses.

If anyone has an idea to fix that…


racke:
Please clean up your pull request (src/lib/stamp-man1, src/sbin/sympa).

src/lib/stamp-man1
```diff
@@ -0,0 +1 @@
+timestamp
```
ikedas, line 1: This should not be included in the source repo.

src/sbin/sympa
```diff
@@ -0,0 +1,912 @@
+#!/usr/bin/perl
```
ikedas, line 1: This shouldn't be included in the repo (`sympa.pl.in` suffices).

ldidry: Sorry, I thought I didn’t added the files, but… my mistake :shrug:

PR #1327, opened 2022-01-19
WWSympa: "quiet" parameter for "add"/"subscribe" requests weren't handled correctly

### LINKED ISSUE #1326: Sympa 6.2.67b.3 build failure
Version
-------

Sympa 6.2.67b.3 from https://github.com/sympa-community/sympa/releases/download/6.2.67b.3/sympa-6.2.67b.3.tar.gz

Installation method
-------------------

My own rpm, derived from official RHEL 7 rpm

Expected behavior
-----------------

6.2.67b.3 should build

Actual behavior
---------------

"make check" fails with:

```
Test Summary Report
-------------------
t/Request_Handler_add+del.t (Wstat: 512 Tests: 1 Failed: 0)
  Non-zero exit status: 2
  Parse errors: No plan found in TAP output
Files=24, Tests=1205, 58 wallclock secs ( 0.19 usr  0.02 sys + 48.61 cusr  5.04 csys = 53.86 CPU)
Result: FAIL
Failed 1/24 test programs. 0/1205 subtests failed.
```

It is possible that this is a stupid error on my part and I am just missing a dependency. However I haven't run into this sort of build failure before. Maybe the following is significant?

https://github.com/sympa-community/sympa/pull/1299

That is the only mention of "make check" on https://github.com/sympa-community/sympa/blob/6.2.67b.3/NEWS.md

ikedas:
- `quiet_subscription` parameter overrided `quiet` option for `add` request (fix #1326)
  - WWSympa: Adding subscribers, sometimes `quiet` parameter was omitted, because template didn't support it.

### DIFF
--- a/src/lib/Sympa/Request/Handler/subscribe.pm
+++ b/src/lib/Sympa/Request/Handler/subscribe.pm
@@ -119,7 +119,7 @@ sub _twist {
     $user->save;
 
     ## Now send the welcome file to the user
-    $request->{quiet} = ($Conf::Conf{'quiet_subscription'} eq "on")
+    $request->{quiet} ||= ($Conf::Conf{'quiet_subscription'} eq "on")
         if $Conf::Conf{'quiet_subscription'} ne "optional";
     unless ($request->{quiet}) {
         unless ($list->send_probe_to_user('welcome', $email)) {
--- a/src/lib/Sympa/Request/Handler/subscribe.pm
+++ b/src/lib/Sympa/Request/Handler/subscribe.pm
@@ -118,9 +118,7 @@ sub _twist {
     $user->lang($list->{'admin'}{'lang'}) unless $user->lang;
     $user->save;
 
-    ## Now send the welcome file to the user
-    $request->{quiet} ||= ($Conf::Conf{'quiet_subscription'} eq "on")
-        if $Conf::Conf{'quiet_subscription'} ne "optional";
+    # Now send the welcome file to the user.
     unless ($request->{quiet}) {
         unless ($list->send_probe_to_user('welcome', $email)) {
             $log->syslog('notice', 'Unable to send "welcome" probe to %s',
--- a/.github/workflows/make-check.yml
+++ b/.github/workflows/make-check.yml
@@ -71,7 +71,7 @@ jobs:
             export LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu
           fi
           autoreconf -i
-          ./configure
+          ./configure --with-defaultdir=`pwd`/default
           cd src; make; cd ..
           make check-local TEST_FILES='xt/perltidy.t' || true
           make check-local
--- a/default/web_tt2/confirm_action.tt2
+++ b/default/web_tt2/confirm_action.tt2
@@ -383,17 +383,20 @@
         [% FOREACH e = email ~%]
             <input type="hidden" name="email" value="[% e %]" />
         [%~ END %]
-        [% IF conf.quiet_subscription == 'optional' %]
-            <div>
-                <input type="checkbox" id="quiet" type="checkbox" name="quiet" value="1"
-                       [% IF quiet %]checked="checked"[%END%] />
-                <label for="quiet">[%|loc%]Quiet (don't send welcome email)[%END%]</label>
-            </div>
-        [%~ END %]
+        <div>
+            <input type="checkbox" id="quiet" name="quiet"
+             value="1"[% IF quiet %] checked="checked"[%END%] />
+            <label for="quiet">[%|loc%]Quiet (don't send welcome email)[%END%]</label>
+        </div>
     [%~ ELSIF confirm_action == 'add_frommod' ~%]
         [% FOREACH i = id ~%]
             <input type="hidden" name="id" value="[% i %]" />
         [%~ END %]
+        <div>
+            <input type="checkbox" id="quiet" name="quiet"
+             value="1"[% IF quiet %] checked="checked"[%END%] />
+            <label for="quiet">[%|loc%]Quiet (don't send welcome email)[%END%]</label>
+        </div>
     [%~ ELSIF confirm_action == 'arc' ~%]
         <input type="hidden" name="month" value="[% month %]" />
         <input type="hidden" name="arc_file" value="[% arc_file %]" />
@@ -536,6 +539,7 @@
     [%~ ELSIF confirm_action == 'subscribe' ~%]
         <input type="hidden" name="email" value="[% email %]" />
         <input type="hidden" name="gecos" value="[% gecos %]" />
+        <input type="hidden" name="quiet" value="[% quiet %]" />
         [% FOREACH i = custom_attribute ~%]
             <input type="hidden" name="custom_attribute.[% i.key %]"
              value="[% i.value.value.replace('\n', '&#10;') %]" />
--- a/default/web_tt2/import.tt2
+++ b/default/web_tt2/import.tt2
@@ -10,9 +10,11 @@
     </div>
     <div>
         [% IF conf.quiet_subscription == 'optional' %]
-            <input id="quiet" type="checkbox" name="quiet" value="quiet"
-                   [% IF quiet %] checked="checked"[%END%] />
+            <input id="quiet" type="checkbox" name="quiet"
+             value="1"[% IF quiet %] checked="checked"[%END%] />
             <label for="quiet">[%|loc%]quiet[%END%]</label>
+        [%~ ELSIF conf.quiet_subscription == 'on' ~%]
+            <input type="hidden" name="quiet" value="1" />
         [% END ~%]
         <input class="MainMenuLinks" type="submit" name="action_import"
                value="[%|loc%]Add subscribers[%END%]" />
--- a/default/web_tt2/review.tt2
+++ b/default/web_tt2/review.tt2
@@ -92,11 +92,14 @@
                 </div>
                 [% IF conf.quiet_subscription == 'optional' %]
                     <div>
-                        <input id="quietly" type="checkbox" name="quiet" />
+                        <input id="quietly" type="checkbox" name="quiet"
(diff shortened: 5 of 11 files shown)


src/lib/Sympa/Request/Handler/subscribe.pm
```diff
@@ -118,9 +118,7 @@ sub _twist {
     $user->lang($list->{'admin'}{'lang'}) unless $user->lang;
     $user->save;
 
-    ## Now send the welcome file to the user
-    $request->{quiet} = ($Conf::Conf{'quiet_subscription'} eq "on")
-        if $Conf::Conf{'quiet_subscription'} ne "optional";
```
ldidry, line 123: I think we should keep that: deporting the `quiet` parameter in the templates, even if it’s hidden, allows users to bypass the `quiet_subscription` global parameter by editing the HTML code of the page.

Same thing in `src/lib/Sympa/Request/Handler/add.pm`

ikedas: See the commit c732ce9 .  When `quiet_subscription` is `optional`, it needs to be handled in the template.  Thus, when it is `on`, it may also be handled in the template. In the end, I decided that the lines in Perl sources were not necessary.

ikedas: Correction: See commit 93dc8b7 .

ldidry: I understand that some things need to be handled in templates, but settings that should not be overriden by users should always be enforced in the backend otherwise users may bypass those settings.

ldidry: I think that removing the check in the backend will also allow to bypass the setting when using mail commands 🤔

ikedas:
I see.  I understood what `quiet_subscription` parameter aims.: By default (with value `optional`), it performs the previous action, i.e. `quiet` parameter is respected. Otherwise (either 'on' or 'off'), `quiet` parameter is overridden.

- c732ce9 :  This "fix" was misguided.  he problem was that default values of parameters were not handled appropriately.
- 93dc8b7 : Though part of changes by this commit may be useful, these would be better to be postponed for more review.

I'll submit another PR including corrected changes of c732ce9.  93dc8b7 will be put aside for a moment.

This PR will be closed.

ikedas:
@ldidry , can you check commit 93dc8b7 to see if it contains the necessary changes?

ldidry:
In https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4, you remove the backend check of the configuration in `src/lib/Sympa/Request/Handler/subscribe.pm` and `src/lib/Sympa/Request/Handler/add.pm`:

```
$request->{quiet} = ($Conf::Conf{'quiet_subscription'} eq "on")
        if $Conf::Conf{'quiet_subscription'} ne "optional";
```

I think you should leave it the way it was. The other modifications are good.

ikedas:
> In [93dc8b7](https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4), you remove the backend check of the configuration in `src/lib/Sympa/Request/Handler/subscribe.pm` and `src/lib/Sympa/Request/Handler/add.pm`:
> 
> ```
> $request->{quiet} = ($Conf::Conf{'quiet_subscription'} eq "on")
>         if $Conf::Conf{'quiet_subscription'} ne "optional";
> ```
> 
> I think you should leave it the way it was. The other modifications are good.

Yes, 93dc8b7 was withdrawn and those lines were left.

How about changes on templates in this commit?

ldidry:
- <https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4#diff-50321a4faef6cfdcb87d4b291470de15db64915ba168b93091a75b9e60f4cc98R386>, I think you should keep the `IF conf.quiet_subscription == 'optional'`: as the `quiet` param is forced in the backend if it’s not optional, there’s no point to show the checkbox to the user, which make him believe he can change that behavior.
- <https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4#diff-50321a4faef6cfdcb87d4b291470de15db64915ba168b93091a75b9e60f4cc98R395>: you should add an `IF conf.quiet_subscription == 'optional'`, same reason
- <https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4#diff-2467e8e19b16c86abf4fe422cfeca4e1111452dd3d65ee85da73a6db45deb7afR16> and other similar situations: I see no point of adding an hidden quiet parameter if we force the param in the backend. If we need to keep the `quiet` param in the templates even when it’s handled in the backend, we can keep it, it’s not a big deal.
We could even, to have the `quiet` param even if `quiet_subscription` is set to `off`, do something like that:

```perl
[% IF conf.quiet_subscription == 'optional' %]
    <input id="quiet" type="checkbox" name="quiet"
        value="1"[% IF quiet %] checked="checked"[%END%] />
    <label for="quiet">[%|loc%]quiet[%END%]</label>
[%~ ELSE ~%]
    <input type="hidden" name="quiet"
        value="[% (conf.quiet_subscription == 'on') ? 1 : 0 %]" />
[% END ~%]
```

To summarise my opinion:
- keep the `IF conf.quiet_subscription == 'optional'`
- if you need to keep the `quiet` param in the templates even if the configuration does not allow users to choose it, make it hidden
- you may want to keep the `quiet` param in the templates even if the configuration says it won’t be quiet

ikedas:
> * [93dc8b7#diff-50321a4faef6cfdcb87d4b291470de15db64915ba168b93091a75b9e60f4cc98R386](https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4#diff-50321a4faef6cfdcb87d4b291470de15db64915ba168b93091a75b9e60f4cc98R386), I think you should keep the `IF conf.quiet_subscription == 'optional'`: as the `quiet` param is forced in the backend if it’s not optional, there’s no point to show the checkbox to the user, which make him believe he can change that behavior.
> * [93dc8b7#diff-50321a4faef6cfdcb87d4b291470de15db64915ba168b93091a75b9e60f4cc98R395](https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4#diff-50321a4faef6cfdcb87d4b291470de15db64915ba168b93091a75b9e60f4cc98R395): you should add an `IF conf.quiet_subscription == 'optional'`, same reason

These are the modifications in `confirm_action.tt2`. I.e. whether `quiet` option should be `on` or `off` already has been decided at the previous steps. Thus, it is basically not necessary to make decision again. OTOH, if we dare to make the decision again, there is no reason to limit it to the `optional` case.

> * [93dc8b7#diff-2467e8e19b16c86abf4fe422cfeca4e1111452dd3d65ee85da73a6db45deb7afR16](https://github.com/sympa-community/sympa/commit/93dc8b7c36c4b194104b25d1bdfa0628c8f78da4#diff-2467e8e19b16c86abf4fe422cfeca4e1111452dd3d65ee85da73a6db45deb7afR16) and other similar situations: I see no point of adding an hidden quiet parameter if we force the param in the backend. If we need to keep the `quiet` param in the templates even when it’s handled in the backend, we can keep it, it’s not a big deal.

I agree. Hidden `quiet` parameters are not necessary at all.

>   We could even, to have the `quiet` param even if `quiet_subscription` is set to `off`, do something like that:
> 
> ```perl
> [% IF conf.quiet_subscription == 'optional' %]
>     <input id="quiet" type="checkbox" name="quiet"
>         value="1"[% IF quiet %] checked="checked"[%END%] />
>     <label for="quiet">[%|loc%]quiet[%END%]</label>
> [%~ ELSE ~%]
>     <input type="hidden" name="quiet"
>         value="[% (conf.quiet_subscription == 'on') ? 1 : 0 %]" />
> [% END ~%]
> ```
> 
> To summarise my opinion:
> 
> * keep the `IF conf.quiet_subscription == 'optional'`
> * if you need to keep the `quiet` param in the templates even if the configuration does not allow users to choose it, make it hidden
> * you may want to keep the `quiet` param in the templates even if the configuration says it won’t be quiet

ikedas:
@ldidry , I rethought.

>   We could even, to have the `quiet` param even if `quiet_subscription` is set to `off`, do something like that:
> 
> ```perl
> [% IF conf.quiet_subscription == 'optional' %]
>     <input id="quiet" type="checkbox" name="quiet"
>         value="1"[% IF quiet %] checked="checked"[%END%] />
>     <label for="quiet">[%|loc%]quiet[%END%]</label>
> [%~ ELSE ~%]
>     <input type="hidden" name="quiet"
>         value="[% (conf.quiet_subscription == 'on') ? 1 : 0 %]" />
> [% END ~%]
> ```
> 
> To summarise my opinion:
> 
> * keep the `IF conf.quiet_subscription == 'optional'`
> * if you need to keep the `quiet` param in the templates even if the configuration does not allow users to choose it, make it hidden
> * you may want to keep the `quiet` param in the templates even if the configuration says it won’t be quiet

I understand that you are planning to do the following for each value of `quiet_subscription`:
  * `on`: `quiet` parameter is forced to be true.
  * `off`: `quiet` parameter is forced to be false.
  * `optional`: `quiet` parameter may be chosen by the users.

Therefore, there is no need to use form widgets (`<input>` tags) to handle `on` and `off` cases. Because they are always forced in `add`/`subscribe` handler.  Now my opinion is:

  * Keep the `IF conf.quiet_subscription == 'optional'`
  * No need for things like hidden tags to handle `on` and `off` cases.

``` code 
[% IF conf.quiet_subscription == 'optional' %]
    <input id="quiet" type="checkbox" name="quiet"
        value="1"[% IF quiet %] checked="checked"[%END%] />
    <label for="quiet">[%|loc%]quiet[%END%]</label>
[% END ~%]
```

ldidry:
Yes, that was exactly that :slightly_smiling_face:

PR #1328, opened 2022-01-19
🚸 — Add anchors to prefs to ease linking (like in FAQ, etc.)

ldidry:
+ change h4 tags to h3, since it doesn’t semantically make sense


racke:
Makes sense to me :+1:

PR #1343, opened 2022-02-03
Fix #533 — Prevent the use of the list address as subscriber

### LINKED ISSUE #533: Prevent the use of the list address as subscriber that receive mails
Some users add the list as subscriber of the list with mail reception, causing mail loops.

Version
-------
6.2.36

Installation method
-------------------
source package

Expected behavior
-----------------
The list should not be able so be added as subscriber or at least should not receive mails from the list.

Actual behavior
---------------
The list can be added to itself as subscriber with mail reception enabled.

ldidry:
In #533, I originally wanted to prevent adding the list address as subscriber with a reception mode which sends mail to the subscriber but:
- it’s far more easy to just prevint adding the list address as subscriber
- it would need to prevent changing the reception mode after subscribing, so more work
- I can’t see a usecase where you want to subscribe with the address of the list. But I think it’s better to have a backend check to prevent it (to avoid typo and other user mistakes)

`po/sympa/sympa.pot` needs to be updated as I introduce a new localized string. Shoud I run `make -C po/sympa sympa.pot-update` and commit `po/sympa/sympa.pot` or is there another workflow for that?

### DIFF
--- a/src/lib/Sympa/List.pm
+++ b/src/lib/Sympa/List.pm
@@ -3126,6 +3126,11 @@ sub add_list_member {
                 $u->{email});
             next;
         }
+        if ($who eq $self->get_id) {
+            $log->syslog('err', 'Ignoring %s which is the address of the list',
+                $who);
+            next;
+        }
         unless (
             $current_list_members_count < $self->{'admin'}{'max_list_members'}
             || $self->{'admin'}{'max_list_members'} == 0) {
--- a/default/mail_tt2/report.tt2
+++ b/default/mail_tt2/report.tt2
@@ -663,6 +663,9 @@ Warning: this message may already have been sent by one of the list's moderators
 [%~ ELSIF report_entry == 'blocklisted_domain' ~%]
   [%|loc(report_param.email || report_param.value)%]Address "%1" belongs to a blocklisted domain[%END%]
 
+[%~ ELSIF report_entry == 'email_is_the_list' ~%]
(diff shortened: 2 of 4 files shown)


ikedas:
> `po/sympa/sympa.pot` needs to be updated as I introduce a new localized string. Shoud I run `make -C po/sympa sympa.pot-update` and commit `po/sympa/sympa.pot` or is there another workflow for that?

I documented the procedure to release new source tarball: See "[How to prepare a new source tarball](https://github.com/sympa-community/sympa/blob/sympa-6.2/support/README.support.md#how-to-prepare-a-new-source-tarball)". As it includes a step to update `sympa.pot` etc., any beta releases may contain the latest catalogs.

ldidry:
Thank you. I updated the [wiki](https://github.com/sympa-community/sympa/wiki)

src/lib/Sympa/List.pm
```diff
@@ -3126,6 +3126,11 @@ sub add_list_member {
                 $u->{email});
             next;
         }
+        if ($who eq $self->get_id) {
+            $log->syslog('err', 'Ignoring %s which is the address of the list',
+                $who);
+            next;
+        }
```
ikedas, line 3133: I feel the codes for error handling in `add.pm` & `subscribe.pm` are redundant.

We'd be better to insert
``` perl
push @$stash_ref, ['user', 'email_is_the_list', {email => $who}];
```
before the `next` line above and withdraw the changes below.

ikedas: ping @ldidry ?

PR #1409, opened 2022-06-03
The owner sees the pop-up stating a message is waiting for moderation even if moderators are defined separately (#1406)

### LINKED ISSUE #1406: The owner sees the pop-up stating a message is waiting for moderation even if moderators are defined separately
6.2.68
-------

Installation method
-------------------
Cent7, mariadb, postfix, rpm

Expected behavior
-----------------
The left menu should reflect a message is waiting for moderation (Moderate > Message (N)).  

Actual behavior
---------------
The web UI reflects a popup window stating a message is awaiting moderation but the left menu is blank and does NOT reflect a Message is awaiting moderation.     I also noticed the 'Message (0)' is missing after one changes the 'who can send message' to a moderated scenari.  See the attached screen shots.

Steps to reproduce
------------------
1. create a new list, note the left menu reflects Moderate with three options (Messages, Subscriptions, Unsubscriptions)
2. change the list to a moderate 'who can send messages' scenario.   
3. send an email to the new list address
4. log into web ui, goto the list home page and a popup display stating there is a message awaiting moderation. 
5. click on the Moderate option in the left menu.  The left menu only displays (Subscriptions, Unsubscriptions) and the Messages option is missing.

Additional information
----------------------
The functionality of a Moderated works as designed but the Web UI does not.   
sympa 6 2 moderate UI w-messages
This image shows the Moderate > Message () UI in left menu.

-------------
6 2 moderated message UI bug
This image shows the Moderate > Message UI is missing from the left menu.  The 'who can send message' scenari is "Private, moderate for non subscribers.

---- scenario -----
title.gettext Private, moderated for non subscribers

is_subscriber([listname],[sender]) smtp,dkim,md5,smime    -> do_it
is_editor([listname],[sender])     smtp,dkim,md5,smime    -> do_it
true()                             smtp,dkim,md5,smime    -> editorkey

ikedas:
This may fix #1406.

### DIFF
--- a/default/web_tt2/info.tt2
+++ b/default/web_tt2/info.tt2
@@ -17,7 +17,7 @@
     [% IF is_owner || is_editor %]
         <hr />
         <h3>[%|loc%]Administrative Options[%END%]</h3>
-        [% IF is_owner || is_editor %]
+        [% IF is_editor %]
             [% IF mod_message && (mod_message > 0) %]
                 <p class="small-8 small-centered columns alert-box info text-center">
                     [%|loc(mod_message)%]There are %1 message(s) awaiting moderation.[%END%]


tmclaren:
What was the proposed solution?

ikedas:
As I [wrote](https://github.com/sympa-community/sympa/issues/1406#issuecomment-1140177662), the 'owner who does not double moderator should not see the pop-up about "message(s) awiting moderation", as they does not see about such messages in the side menu'.

tmclaren:
This is not a solution.  This approach obfuscates the privileged owners view into the moderation queue when moderators are defined.

ikedas:
As I [wrote](https://github.com/sympa-community/sympa/issues/1406#issuecomment-1142902446), (1) if the owner want to view it, they can give the role of moderator to themselves.  OTOH (2) some owners may not want to view everything by their own.
Currently Sympa's default behavior is (2) and (1) needs optional setting.

If you want to make (1) the default behavior of Sympa, i.e. to make (2) optional, you would be better to submit a separate issue.

ikedas:
@tmclaren , please try it: When you are a owner, please add yourself as a moderator too, then check if the pop-up about moderation is shown.  That's what this PR indended.

tmclaren:
I realize the owner could add themselves to the moderator's list but that has side effects:    
1) they will start receiving the automated messages from the server when a message is waiting for moderation in addition to seeing the queue in the web UI.   Many list owners do not want that extra email event since they've assigned others to moderate.
2) they will see the moderation queue when they log into the UI.

A privileged list owner should not be restricted from seeing the moderation queue when they assigned other persons to moderate the list.   Think about the professor and teaching assistant (TA) relationship:  Prof creates and configures a class list and the TA moderates the lists.  However, the Prof will spot check the list to verify all inbound messages are being moderated in a timely manner.    They should be able to see the moderation queue when the log into the web UI.   Sadly, this change prevents the privileged owner from ever seeing the moderation queue and IMHO has only made this issue worse.    

I still recommend, the privileged list owner get the popup message about pending moderation messages AND the left menu reflect the same information.   In general, I believe the privileged list owner should NOT be restricted from their mail list settings or behavior unless the listmaster has configured the server to specifically restrict the view of a setting/behavior in the  ~/edit_list.conf.

tmclaren:
Unfortunate!  You are obfuscating a fundamental permission for privileged
owners. Hopefully this hack will get documented as a caveat in the owners
doc.   Sad…

PR #1411, opened 2022-06-08
File names in incoming spool can be truncated (#1410)

### LINKED ISSUE #1410: File names in incoming spool can be truncated
Version
-------

All previous versions.

Installation method
-------------------

Any.

Expected behavior
-----------------

The file name in the spool should not be truncated regardless of any conditions.

Actual behavior
---------------

Using fairly long list name, the file name in the spools generated by queue programs (`queue`, `bouncequeue` & `familyqueue`) may be truncated.

Steps to reproduce
------------------

  1. Set up Sympa with [Postfix "virtual domain" setting](https://sympa-community.github.io/manual/install/configure-mail-server-postfix.html#virtual-domain-setting).
  2. Add this line to `transport.sympa`:
     ``` code
     test-address@example.jp sympa:1234567890123456789012345678901234567890123456789012345678901234@6789.123456789.123456789.123456789.123456789.123456789.example.jp
     ```
     then run `postmap transport.sympa`.
  3. send a message to `test-address@example.jp`.
  4. Check the content of spool by running:
      ``ls -R $SPOOLDIR/msg``

The result is as below:
``` bash
# ls -R /var/spool/sympa/msg
/var/spool/sympa/msg:
bad

/var/spool/sympa/msg/bad:
1234567890123456789012345678901234567890123456789012345678901234@6789.123456789.123456789.123456789.123456789.123456789.examp
```
Correct file name should end with ...``.example.jp.ttttttttttt.ppppp`` (`ttt...` is a Unix time, and `ppp...` is a PID number), but the file name is truncated to 125 bytes.

Additional information
----------------------

RFC 5321 [states](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1) that  the maximum total length of e-mail address should be 254 octets (Actual implementations may allow longer addresses).  On Sympa 6.2 or later, maximum length of list address is [limited to 131 octets](https://github.com/sympa-community/historic-sympa/blob/4eef997e9a64c2f13d6d100cca85e014b6e0199b/src/lib/Sympa/Constants.pm.in#L55-L56) by default (On earlier versions there was no significant limit).

The name of the file in Sympa's incoming queue requires additional tens of octets for some information.  However, queue programs have given [only 128 bytes of buffer for file names](https://github.com/sympa-community/historic-sympa/blob/94073fb998dab1acb666db9f2469345a9f59f689/src/queue.c#L10). Thus file names can be truncated even if appropriate list names are used.

ikedas:
This may fix #1410, and makes some cosmetic changes.

### DIFF
--- a/src/libexec/bouncequeue.c
+++ b/src/libexec/bouncequeue.c
@@ -32,7 +32,6 @@
 #include <unistd.h>
 #include <time.h>
 
-static char     qfile[128];
 static char     buf[16384];
 static int      i, fd;
 
@@ -90,18 +89,17 @@ readconf(char *file)
 int
 main(int argn, char **argv)
 {
-   char	*bouncedir;
-   char        *listname;
+   char	*bouncedir, *listname, *qfile;
    int			firstline = 1;
 
    /* Usage : bouncequeue list-name */
    if (argn != 2) {
       exit(EX_USAGE);
    }
-
-   listname = malloc(strlen(argv[1]) + 1);
-   if (listname != NULL)
-     strcpy(listname, argv[1]);
+   if (!*(listname = argv[1]))
+      exit(EX_USAGE);
+   if ((qfile = malloc(strlen(listname) + 43)) == NULL)
+      exit(EX_TEMPFAIL);
 
    if ((bouncedir = readconf(CONFIG)) == NULL)
       exit(EX_CONFIG);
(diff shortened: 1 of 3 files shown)


xavierba:
The magic number 43 should be explained or even made a variable as it it used in several locations.

ikedas:
@xavierba , thanks for advice.  Done!

PR #1421, opened 2022-06-15
[CLI] -f|--config option ignored.

dverdin:
In arrange sub, the configuration is loaded two times. Only the first loading was using the config option value.

[Note] this post has been edited by admin.


ikedas:
This looks insufficient and may be replaced by #1425 .

PR #1425, opened 2022-06-16
CLI: Only the first loading of sympa.conf used the --config option

ikedas:


dverdin:
That works fine on my test server.

PR #1426, opened 2022-06-16
Fix outdated link to Data::Password variables

racke:


ikedas:
Fixed by modifying catalog on translation server: cf. d073fab

racke:
Sorry, but I don't like this solution. The source code is still outdated and that is reflected in the online documentation.

ikedas:
> Sorry, but I don't like this solution. The source code is still outdated and that is reflected in the online documentation.

The source was modified, too (using [correct_msgid](https://github.com/sympa-community/sympa/blob/sympa-6.2/support/README.support.md#correct_msgid)).

However, since this method needs frequent access to the message catalog, general method (letting translations fuzzy and waiting correction by translators) can be reasonable.

ikedas:
Sorry. I was doing maintenance on the translation server over the weekend, so I applied the suggested updates on my side. Thanks for correcting the message.

PR #1431, opened 2022-06-23
Inclusion: log mistakenly notes failure of inclusion

### LINKED ISSUE #1429: Logs reflect Failure when include from SQL is carried out during nosync_time_ranges
Version
-------
6.2.60

Installation method
-------------------
Debian Bullseye Package

Expected behavior
-----------------
We have plenty of lists with subscribers defined in a SQL external database; all these definitions contain nosync_time_ranges. In my understanding, inclusion of subscribers should not be even started during this periods and logs should not collect any entry related to this. But if the desing of sympa does not allow to prevent this inclusion from being started and logs to be collected, then logs should not reflect a Failure since this is not a failure in the process but an regular behaviour. 

Actual behavior
---------------
Logging collects entries like this one:

```
Jun 22 07:05:41 hostname task_manager[1210995]: info Sympa::Request::Handler::include::_twist() Sympa::Request <action=include;context=list_name@domain.com;role=member>: Failure, 0 added, 720 held, 0 updated
```

Steps to reproduce
------------------
Define an include_sql_query with nosync_time_ranges.

```
include_sql_query
name students
db_name expxcm-dsn
db_type ODBC
db_passwd PASSWORD
nosync_time_ranges 06:30-07:45
sql_query SELECT email FROM table WHERE role = 1
db_user USERNAME
```

ikedas:
**Now under self review**

Inclusion: When all of the data sources are skipped by nosync_time_ranges, log notes failure of inclusion.

This may fix #1429 .

### DIFF
--- a/src/lib/Sympa/Request/Handler/include.pm
+++ b/src/lib/Sympa/Request/Handler/include.pm
@@ -194,10 +194,15 @@ sub _twist {
     my %result =
         (added => 0, deleted => 0, updated => 0, kept => 0, held => 0);
     my $succeeded = 0;
+    my $skipped   = 0;
     foreach my $ds (@{$dss || []}) {
         $lock_fh->extend;
 
-        next unless $ds->is_allowed_to_sync;
+        unless ($ds->is_allowed_to_sync) {
+            $skipped++;
+            next;
+        }
+
         my %res = _update_users($ds, $start_time);
         unless (%res) {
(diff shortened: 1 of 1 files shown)


ikedas:
### Known bug

Among multiple data sources, if there is no common time period that is not excluded by `nosync_time_range` in all of them, expired users will never be deleted from the list.

Issue #115, opened 2017-11-01
Probem typo


racke:
```
./default/mail_tt2/user_notification.tt2:52:Please contact the listmaster (%2@%3) to fix this probem before re-posting your message.[% END %]
./po/sympa/sympa.pot:11327:"Please contact the listmaster (%2@%3) to fix this probem before re-posting your message."
```

ikedas:
Also, "You message to list %1 could not be delivered." in the same entry would be "You*r* ...".

ikedas:
Reported typos and corresponding catalog entries were [fixed](https://github.com/sympa-community/sympa/commit/1e2e0941fd771a702b7d04f2166bbe59d90ca6cd#diff-e18b607efb46e4cdee50fb21b43a2e07) (I wrote a script to reflect changes on en_US.po to the sources).

I suppose, to catch complaint about typos, this issue would be better to be left open.

qosobrin:
#: src/lib/Sympa/ConfDef.pm:449
msgid "The number of seconds a bulk worker will remain running without processing a message before it spontaneously exists."

should be 

msgid "The number of seconds a bulk worker will remain running without processing a message before it spontaneously exits." 

or even 

msgid "The number of seconds a bulk worker will remain running without processing a message before it spontaneously quits."

qosobrin:
#: src/lib/Sympa/ConfDef.pm:437
msgid "The minimum number of packets before bulk daemon forks the new worker to increase sending rate."

should be 

msgid "The minimum number of packets before bulk daemon forks a new worker to increase sending rate."

ikedas:
@qosobrin, thanks for reporting typo.

If possible, please add your fix to english catalog:
  - On [English page of translation site](https://translate.sympa.org/en_US/), add correct string as a "translation" of incorrect msgid.

Your change on the site above will be reflected to source distribution in the future release.

qosobrin:
Done. I could only "Suggest" my commits in the translation site; it didn't allow me to "Send"

qosobrin:
By the way, what do you prefer me to do with next strings? Should I publish them here and commit the "translation" later? Or should I commit the translation directly?

qosobrin:
I have a doubt about this string:

#: src/lib/Sympa/ConfDef.pm:694
msgid "Remove subscribers bouncing remind message"

Shouldn't it be "Remove bouncing subscribers remind message"? I read the source and I can not understand the context it is being used.

ikedas:
[...]

PR #1432, opened 2022-06-24
Fix typos

fingolfin:
This fixes a bunch of typos.

You might not want to merge this as-is, as it could break existing
installation with customized templates? But I don't know what your
policies on this are esp. how you prefer to deal with it; so I just
did a global "search and replace", and hope you can give a hint as
to which places (if any) I maybe should not touch.

Or perhaps you don't mind and want to just merge :-).


ikedas:
@fingolfin , thank you for the corrections!

Among your fixes, these would be better to be corrected by submitting PR.

  - Changes of a parameter name in the code:
      - default/web_tt2/review.tt2
      - src/cgi/wwsympa.fcgi.in

    (A suggestion on the fix: For templates customized for earlier versions, it is better to duplicate parameter value for old parameter name as well.)

  - Correction in the source commnent:
      - support/xgettext.pl

The others, typos in the messages/strings to be translated, may be corrected by adding correction on the translation server. --- If possible, please add your fix to english catalog:

  - On [English page of translation site](https://translate.sympa.org/en_US/), add correct string as a "translation" of incorrect msgid.

Your changes on the site above will be reflected to source distribution in the future release.  PO files need not be corrected manually, as they will also be updated accordingly.

racke:
IMHO typos in the source code should be fixed there and not only in the English translation. Otherwise the typos are exposed and diligent contributors will report them again. The approach in #1426 is the way it should be done.

ikedas:
> IMHO typos in the source code should be fixed there and not only in the English translation. Otherwise the typos are exposed and diligent contributors will report them again. The approach in #1426 is the way it should be done.

My approach to #1426 was [to modify _both_ the source code and the message catalog _at once_](https://github.com/sympa-community/sympa/commit/d073fab) based on the single information entered into the translation server (we can do this with the [correct_msgid](https://github.com/sympa-community/sympa/blob/sympa-6.2/support/correct_msgid) script: Please confirm the content of script).  I don't think the pain of another report coming in before above process is complete is much different than the pain of a PR being submitted with the same fixes before one PR has been merged.

Anyways, if you have objection to the method that was used to solve #115, How about discussing it not as an issue about this PR but as a more general release management issue?  In the case of this PR, please let us do it the way we have been doing it.

ikedas:
@fingolfin , sorry for the confusion.  As I'll fix "the others" in [previous comment](https://github.com/sympa-community/sympa/pull/1432#issuecomment-1166073959), could you please submit another PR to fix remainder, parameter names and comment in sources code?  Thank you.

fingolfin:
@ikedas I made PR #1433 but it wasn't clear to me what you want to be in there and what not... so for now I just dropped the changes to `po` files there. Please let me know if I should drop more (and what)

ikedas:
Hi @fingolfin ,

At least po files are prone to conflicts, so you don't have to correct them in the PR.
Further complication would be a burden on you, so I will accept your PR and make other adjustments on my side.

Thank you for your careful corrections!

Issue #1011, opened 2020-10-08
Quiet mode is ignored when (un)subcrubing from SOAP server


ynerant:
I use the SOAP server to automate some registrations from my website.
Creating a list goes successfully, and I can subscribe or unsubscribe members.
However, the notifications that the user joined the list or got unsubscribed are not sent, as if quiet mode was enabled, even if this is not the case.
All is working well through the web server.

Version
-------
6.2.40

Installation method
-------------------
Debian package, from APT, with the SOAP server enabled

Expected behavior
-----------------
When registering someone to a list without the quiet mode through the SOAP server, a notification should be sent.

Actual behavior
---------------
No mail is sent.

Additional information
----------------------
Syslog indicates that the quiet mode is well received by the server: (I anonymized my personal data)

```
Oct  8 22:02:05 machine sympasoap[387]: info Sympa::WWW::SOAP::add(list = test-create@lists.example.com, email = toto@example.com, quiet = false) From toto@example.com via proxy application 
Oct  8 22:04:21 machine sympasoap[388]: info Sympa::WWW::SOAP::add(list = test-create@lists.example.com, email = toto@example.com, quiet = false) From toto@example.com via proxy application 
Oct  8 22:04:46 machine sympasoap[390]: info Sympa::WWW::SOAP::del(list = test-create@lists.example.com, email = toto@example.com, quiet = true) From toto@example.com via proxy application 
Oct  8 22:04:49 machine sympasoap[386]: info Sympa::WWW::SOAP::del(list = test-create@lists.example.com, email = toto@example.com, quiet = true) From toto@example.com via proxy application 
Oct  8 22:04:52 machine sympasoap[389]: info Sympa::WWW::SOAP::add(list = test-create@lists.example.com, email = toto@example.com, quiet = false) From toto@example.com via proxy application 
Oct  8 22:05:25 machine sympasoap[387]: info Sympa::WWW::SOAP::add(list = test-create@lists.example.com, email = toto@example.com, quiet = false) From toto@example.com via proxy application 
Oct  8 22:05:28 machine sympasoap[388]: info Sympa::WWW::SOAP::del(list = test-create@lists.example.com, email = toto@example.com, quiet = true) From toto@example.com via proxy application 
```

ikedas:
Hi @galaxyoyo ,

I guess `quiet` parameter should be given as `1` or `0`, instead of `true` or `false`.

ynerant:
[...]

PR #1456, opened 2022-07-27
Import SOAP encoding schema in WSDL

cgx:
This fix adds the required import of the SOAP encoding schema in WSDL so clients don't have to [deal with a broken schema](https://github.com/suds-community/suds#fixing-broken-schemas).


ikedas:
Hi @cgx.

I guess #1011 is caused by the problem this pr addresses.  Is it right?

ikedas:
Merged.  Thanks you for fixing bug!

PR #1463, opened 2022-08-17
Use new domain name and list addresses for Sympa in the source

ikedas:
See [announcement](https://www.sympa.community/2022/08/16/community-has-a-new-domain-name.html).


racke:
I think that is also a good candidate for 6.2.70. I'm going to give that a thorough review before including it. Do we really need www.sympa.community for the web URL? sympa.community looks much nicer to me :-)

ikedas:
> Do we really need [www.sympa.community](http://www.sympa.community) for the web URL? sympa.community looks much nicer to me :-)

This is a restriction by GitHub Pages.  It ties arbitrary hostname to GitHub with a CNAME record (See [documentation](https://docs.github.com/pages/configuring-a-custom-domain-for-your-github-pages-site)).  E.g., in our case:
``` code
www.sympa.community.    86400   IN       CNAME   sympa-community.github.io.
```
However, since hostname with CNAME record cannot be given other records (see RFTC 2181), the feature above makes other usage of that hostname (e.g., MX) impossible.

As a next best solution, so I made HTTP access to [sympa.community](https://sympa.community) be redirected to www.sympa.community: sympa.community doesn't need CNAME record, so it may also be used for MX.

PR #1496, opened 2022-10-12
✨ login right away when there is only sso configured

m0ppers:
we only use sso for sympa logins and it is really annoying to press the login button (we are deep linking into confidential archive URLs which currently triggers a not authenticated message which users don't understand).

this works for me but I have no idea what I am doing. I never did anything in perl.

would be great to have this (or some other mechanism that solves this) by default.

### DIFF
--- a/src/cgi/wwsympa.fcgi.in
+++ b/src/cgi/wwsympa.fcgi.in
@@ -1432,6 +1432,26 @@ while ($query = Sympa::WWW::FastCGI->new) {
                 $session->{'do_not_use_cas'} = 1
                     unless $param->{'redirect_to'} =~ /http(s)+\:\//i;
             }
+        } elsif (((!$session->{'email'}) || ($session->{'email'} == "nobody")) && scalar @{$auth_services || []} == 1 && ($param->{'sso_number'} == 1)) {
+            # only one auth method specified (sso). try to login right away
+            my $service_id = each %{$param->{'sso'}};
+            my $auth = Conf::get_auth_service(
+                    $robot, 'generic_sso', $service_id
+            );
+            if ((defined $auth->{email_http_header}) && (defined $ENV{$auth->{email_http_header}})) {
+                my @email_list = split(
+                    /$auth->{http_header_value_separator}/,
+                    Sympa::Tools::Text::canonic_email(
+                        $ENV{$auth->{email_http_header}}
+                    )
+                );
+                my $email = $email_list[0];
+                if ($email) {
+                    $param->{'user'}{'email'}    = $email;
+                    $session->{'email'}          = $email;
+                    $session->{'auth'}           = 'generic_sso';
+                }
+            }
         }
 
         if ($param->{'user'}{'email'}) {


ikedas:
Hi @m0ppers , thank you for suggestion.

According to the comment, current generic_sso authentication assumes the following scenario.

https://github.com/sympa-community/sympa/blob/f71f5ce96fc386993b151bff41a83fe205f8e2e8/src/cgi/wwsympa.fcgi.in#L3287-L3301

However, if the authn module recognizes that the user has already authenticated, then the `init` subaction in above should be complete and the user may not need to perform the login operation again.

I will review this PR some more, and ask some others to review it.

salaun-urennes1:
Hello,

The generic_sso had been introduced to enable Shibboleth authentication, see https://www.sympa.community/manual/customize/authentication-web.html and with Shibboleth authentication. With Shibboleth module for Apache authentication workflow is triggered when the user hits a given URL (http://yourhost.yourdomain/sympa/sso_login/inqueue for example). Therefore, with Shibboleth, there is little chance that `$ENV{$auth->{email_http_header}}` before the user clicks the "login" button in Sympa.

I guess you don't use Shibboleth, but another authentication technology ?
I also suppose you have some sort of `require valid-user` directive in your HTTP server to trigger authentication any time the user accesses your Sympa's GUI. Therefore I understand your need.

I don't see any objection for your PR (except that it duplicates existing code...but SSO code in `wwsmpa.fcgi` is already a nightmare).

However I don't have a Sympa instance with `generic_sso` enabled to test your code.

m0ppers:
hello. yes I am using the openid-connect apache module and it works very well:

This is my apache config

```
    OIDCProviderMetadataURL http://keycloak.example/auth/realms/users/.well-known/openid-configuration
    OIDCClientID sympa
    OIDCClientSecret super-secret

    # OIDCRedirectURI is a vanity URL that must point to a path protected by this module but must NOT point to any content
    OIDCRedirectURI https://sympa.example/oidc/redirect_uri
    OIDCCryptoPassphrase passphrase
    OIDCRemoteUserClaim email
    OIDCScope "openid profile email"

    <Location /static-sympa>
      AuthType openid-connect
      Require valid-user
    </Location>

    <Location /oidc/redirect_uri>
      AuthType openid-connect
      Require valid-user
    </Location>

    <Location /sympa>
      SetHandler "proxy:unix:/var/run/wwsympa/wwsympa.socket|fcgi://localhost"
      AuthType openid-connect
      Require valid-user
    </Location>
```

This requires the users to be authenticated via keycloak before they can fetch anything from sympa.

Once it hits sympa the sympa authentication with my changes kicks in and the user is logged in right away.

My auth.conf looks like this:

```
generic_sso
    service_name               keycloak
    service_id                  keycloak
    http_header_list            OIDC_CLAIM_email
    email_http_header           OIDC_CLAIM_email
```

that's it. note that the http header is case sensitive (to my knowledge http headers should always be processed case insensitive but this is another problem).

salaun-urennes1:
@m0ppers Thanks for providing these details; it helps understand your context.

One option to improve your user experience without changing Sympa's code would be to change your Apache config as follows :
```
<Location /sympa/sso_login/keycloak>
      SetHandler "proxy:unix:/var/run/wwsympa/wwsympa.socket|fcgi://localhost"
      AuthType openid-connect
      Require valid-user
</Location>
```
This would let users navigate in Sympa's GUI without needing systematic authentication. The user would click on the login button when required only. This would be more aligned with Sympa's web interface philosophy (allowing anonymous navigation) and with the `generic_sso` design.

m0ppers:
I see what you mean but users don't even access the sympa interface at all normally :scream: 

we completely automated mailinglist handling (namely there are lists for groups and projects in our system that are automatically created and users will also be subscribed automatically to them based on some rules). the only real feature we are using (or want to use) from the UI is the web archive. we have an external search which can search several sources (sympa mailinglist archives being one of them) and then should be able to deep link into the sympa web archive.

ikedas:
@m0ppers & @salaun-urennes1 ,

So `generic_sso` mechanism assumes some rich functionality. Then why not add another authentication mechanism that simply uses the authentication information received from the web server's authentication module?  Suppose we name it  "cgi" mechanism in `auth.conf`:
``` code
cgi
    auth_type openid-connect
    remote_user_variable REMOTE_USER
    regexp @our[.]domain[.]org
```
If specified authn module (determined by AuthType) passed the information of authenticated user (typically using REMOTE_USER CGI variable), Sympa uses it as the user email.  Otherwise the user is considered anonymous.

salaun-urennes1:
@ikedas Your proposition makes sense, because the `generic_sso` mechanism is probably over-sophisticated for many Sympa admin's needs. Can you explain what would `auth_type` and `regexp` parameters would be used for?

ikedas:
Just a thought and not actually tried, but in my thought: Sympa allows only the authentication scheme ("Basic", "Digest", "openid-connect" etc., set by AuthType directive on Apache) specified by `auth_type` parameter, and uses the value of REMOTE_USER variable (this is the standard variable by CGI specification) as authenticated user.

m0ppers:
that would work for me

ikedas:
> The only situation where it might be useful would be if multiple authentication modules are enabled on the Apache server. Right?
Yes, I think so.  `auth_type` parameter in `auth.conf` may be optional.

I will look into this idea this weekend to see if it is feasible.

ikedas:
I submitted #1497 to deal with this issue.

ikedas:
Please check the PR #1499.
And please continue discussion about this issue on #1497.

m0ppers:
superseded by #1499

helendobb10 (commented):
I need cash

PR #1513, opened 2022-11-12
Fix SQLite upgrade with lowercase types

k0lter:
It seems that "occasionally" the SQL schema types in SQLite are not lower case so the upgrade in this case fail.

This patch was submitted in the following bug report: https://bugs.debian.org/1022943

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## [6.2.70](https://github.com/sympa-community/sympa/tree/6.2.70) (2022-10-09)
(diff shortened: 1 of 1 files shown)


k0lter:
It seems that I've mixed up my 2 PR...

What do you think, if I "drop" them and create them again properly?

ikedas:
> It seems that I've mixed up my 2 PR...
> 
> What do you think, if I "drop" them and create them again properly?

Yes.  You made unrelated changes on the same branch in your fork.

To submit a PR, you'd be better to create a "topic branch" by each PR in your local clone, then make changes on that branch and push it into your fork.  Using "main branch" (currently `sympa-6.2`) to submit PR seems not a good idea.

PR #1516, opened 2022-11-12
Fix SQLite upgrade with lowercase types

k0lter:
It seems that "occasionally" the SQL schema types in SQLite are not lower case so the upgrade in this case fail.

This patch was submitted in the following bug report: https://bugs.debian.org/1022943


ikedas:
Merged.  Thanks for fix the bug!

PR #1518, opened 2022-11-13
Fix 'sympa config' command return code to be 0 when there are no changes

k0lter:
While trying to update a config value using `sympa config` command when there are no changes, it die() with a return code > 0.

It means that the command is not idempotent.

As an example, we started to use it in Debian Maintainer scripts with some config parameters instead of our custom routines to update configuration.

### DIFF
--- a/src/lib/Sympa/CLI/config.pm
+++ b/src/lib/Sympa/CLI/config.pm
@@ -65,7 +65,7 @@ sub _run {
             only_changed => 1,
             filter => ([@{$options->{output} // []}, qw(explicit mandatory)])
         );
-        die "Not changed.\n" unless defined $out;
+        return 1 unless defined $out;
 
         my $sympa_conf = Sympa::Constants::CONFIG();
         my $date = POSIX::strftime('%Y%m%d%H%M%S', localtime time);


ikedas:
Hi @k0lter ,

It makes sense not to die(), however I feel the message "Not changed." helps.

k0lter:
Hi @ikedas 

It will pollute stdout especially when used multiple times in scripts.

Could you consider adding a `-q` (`--quiet`) argument (disabled by default)?

Thanks,

ikedas:
As `--quiet` has already been used by the other commands, I suggest `-s` / `--silent`.
``` diff
-use constant _options   => qw(output|o=s@);
+use constant _options   => qw(output|o=s@ silent|s);
```
(BTW since stdout pollution is the problem for all commands, I'd like to make this option available for all the other commands in the future.)

k0lter:
Sounds good to me. Thanks.

racke:
Yes, but I would suggest to use "--noout" or something similar as it is easy to confuse --quiet and --silent.

ikedas:
`--quiet` is used not to notify the target user, but `--silent` is certainly confusing.

`--noout` is used by something including openssl and xmllint. I'm ok.

ikedas:
Though I've not tested yet, I've added the `--noout` option.

https://github.com/sympa-community/sympa/compare/sympa-6.2...ikedas:sympa:k0lter/sympa-config-command-rc

k0lter:
Thanks @ikedas

racke:
> Though I've not tested yet, I've added the `--noout` option.
> 
> [sympa-6.2...ikedas:sympa:k0lter/sympa-config-command-rc](https://github.com/sympa-community/sympa/compare/sympa-6.2...ikedas:sympa:k0lter/sympa-config-command-rc)

"Skip output" is easier to parse for the human brain than "No output".

ikedas:
Thanks racke.  And re-pushed correction and more changes.

PR #1533, opened 2022-11-27
CLI: Allow hyphens in options

ikedas:
CLI: Both hyphens and underscores are allowed in command line options (except for the prefix hyphen(s)).


racke:
Finally :+1: . Thanks a lot!

PR #1534, opened 2022-11-28
Fixing SQL custom attributes synchronization.

dverdin:
Currently, SQL custom attributes fails because:

  - it calls Sympa::Tools::Data::encode_custom_attributes with the wrong data structures for the custom attributes,
  - print is called instead of printf.

All this is fixed here.


ikedas:
Insufficient fix.  Close.

dverdin:
And hello to you too.
Good manners are never lost.

PR #1550, opened 2022-12-13
Skip initial notification email for new user with nomail reception

### LINKED ISSUE #1537: New list owner receives email even though no notifications was checked
The idea of this check box is that the list owner doesn't receive notifications about the list. It would be logically if the owner doesn't receive the initial email "Management of list ..."

Version
-------
6.2.70

Installation method
-------------------
From source.

Expected behavior
-----------------
Owner doesn't receive emails for his list.

Actual behavior
---------------
Sympa send an email with the subject "Management of list ..." to the new owner.

Steps to reproduce
------------------
Add owner to the list with "no notifications" checked.

racke:
This may fix #1537. In addition it breaks current behavior so some users might be surprised.

### DIFF
--- a/src/lib/Sympa/Request/Handler/add.pm
+++ b/src/lib/Sympa/Request/Handler/add.pm
@@ -177,7 +177,7 @@ sub _report_user {
     my $sender = $request->{sender};
 
     # Notify the new list owner/editor
-    unless ($request->{quiet}) {
+    unless ($request->{quiet} or $request->{reception} eq 'nomail') {
         Sympa::send_notify_to_user($list, 'added_as_listadmin', $email,
             {admin_type => $role, delegator => $sender});
         $self->add_stash($request, 'notice', 'user_notified',


src/lib/Sympa/Request/Handler/add.pm
```diff
@@ -177,7 +177,7 @@ sub _report_user {
     my $sender = $request->{sender};
 
     # Notify the new list owner/editor
-    unless ($request->{quiet}) {
+    unless ($request->{quiet} or $request->{reception} eq 'nomail') {
```
ikedas, line 180: If `$request->{reception}` was not given, it'd be better to consider `default_owner_options.reception` or `default_editor_options.reception` parameter.

racke: Thanks for your advice @ikedas

ikedas (approved):

PR #1558, opened 2022-12-22
Replace obsolete Ansible role in README

racke:


ikedas:
How about swapping current repository with yours? In other words, how about moving `ansible-role-sympa` to sympa-community?

racke:
Yes, I can certainly move ansible-role-sympa to sympa-community. I would like to retain the name ansible-role-sympa though.

ikedas:
I think the name ansible-role-sympa is fine.

racke:
Moved the repository to sympa-community. My next step would be to add a notice to the obsolete repository and archive that. Is that OK @ikedas

ikedas:
Okey!

ikedas (approved):

PR #1574, opened 2023-01-06
Issue 1573 Archived crash with html sanitizer on regex recursion

DLalot:
Just catching the exception to avoid looping (400 crashed mails received just for one bad email)
It's a very basic solution to consider that problem as a simple archive problem.
Should be better to just put it in spool bad without annoying the listmaster. But  it's a little bit too difficult for me.

### DIFF
--- a/src/lib/Sympa/Spindle/ProcessArchive.pm
+++ b/src/lib/Sympa/Spindle/ProcessArchive.pm
@@ -460,10 +460,11 @@ sub _mail2arc {
             }
         }
     }
-
-    unless ($archive->store($message) and $archive->html_store($message)) {
-        return undef;
-    }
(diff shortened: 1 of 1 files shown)


DLalot:
With the patch, sympa just said that something went wrong. But I believe that this phishing could be distributed in tons of lists. Should we say something ? From my point of view, we should just log it but without sending emails to listmaster..
 
Sympa n'a pas pu archiver le message
'1672985091.1672985097.211337.communication@xx.fr,1389,8655'.
Celui-ci est déplacé dans le répertoire '/home/sympa/spool/outgoing/bad'.
Consultez les logs pour plus de détails.

ikedas:
Superseded by #1575

PR #1584, opened 2023-01-25
When owners/moderators are added, "N subscribers added" is shown

ikedas:
Version: 6.2.68 and later.


racke:
I noticed that too. Thanks for fixing it!

racke (approved):

PR #1648, opened 2023-03-21
Add autofocus to input text web

dominiquefournier:


ikedas:
Hi @dominiquefournier ,

Could you please combine multiple PRs into one?

---
BTW, IMHO we'd be better to [consider accessibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#accessibility_considerations). Is there a good way to deal with this?

dominiquefournier:
@ikedas I don't know how combine multiple PR into one. Do you have a documentation ?

ikedas:
You'd be better to re-submit a new PR including multiple commits.

ikedas:
Additionally, how about adding autofocus attribute to these as well?
-   login.tt2
-   renewpasswd.tt2

PR #1659, opened 2023-04-21
In edit_list_request page, sort scenario on titles

salaun-urennes1:
Currently scenario in edit_list_request pages are sorted on scenario filenames.
As reported in https://lists.sympa.community/msg/fr/2023-04/1TZ5Q-_qRxSqJkG0zscNjQ, it would make sense to sort them on scenario titles.

This PR does the sort in TT2 file

### DIFF
--- a/default/web_tt2/config_common.tt2
+++ b/default/web_tt2/config_common.tt2
@@ -154,20 +154,22 @@
     [% IF pitem.privilege == 'write' ~%]
         <select name="single_param.[% ppaths.join('.') %].name"
                 id="param.[% ppaths.join('.') %]">
-            [% FOREACH scenario = pitem.format ~%]
-                <option value="[% scenario.value.name %]"
-                        [%~ IF scenario.value.name == val.name %] selected="selected"[% END %]>
-                    [%~ scenario.value.title %]
-                    [%~ IF is_listmaster %] ([% scenario.value.name %])[% END ~%]
+			[% SET all = pitem.format.values %]
+            [% FOREACH scenario = all.sort('title') ~%]
+                <option value="[% scenario.name %]"
+                        [%~ IF scenario.name == val.name %] selected="selected"[% END %]>
+                    [%~ scenario.title %]
+                    [%~ IF is_listmaster %] ([% scenario.name %])[% END ~%]
                 </option>
             [% END %]
         </select>
     [%~ ELSE ~%]
-        [% FOREACH scenario = pitem.format ~%]
-            [% IF scenario.value.name == val.name ~%]
-                [% scenario.value.title %]
+		[% SET all = pitem.format.values %]
+        [% FOREACH scenario = all.sort('title') ~%]
+            [% IF scenario.name == val.name ~%]
+                [% scenario.title %]
                 [%~ IF is_listmaster %]
-                    ([% scenario.value.name %])
+                    ([% scenario.name %])
                 [% END %]
             [%~ END %]
         [%~ END %]


ikedas:
I have an objection.

Sorting text in descending order by character code does not always result in an intuitively acceptable order (this is especially true for some languages including Japanese and Chinese).

Furthermore, changing the order of items according to language may cause embarrassment during user support. --- "Select the first item in the dropdown list. ...Yes. If the item is written in English, it is the first item".

salaun-urennes1:
Would it make sense to have a configuration parameter to control this behavior, thus responding to the need of european users, while leaving the freedom for others to keep the current behavior?

ikedas:
However, sorting by character code may not necessarily be convenient for many European users either. IMO the "&agrave;" seems placed after the "z".

salaun-urennes1:
That's true, alphabetical sort is not perfect, but still better than the current situation.

Here is an example on our Sympa main server : 
image

ikedas:
That's just why that English and French share many of the same word roots, so the sort order is often the same. In other words, your language is too specific.  Check by changing the language setting to Russian, Finnish, Magyar and so on.

If we want to allow users to choose the order they prefer, I think, we need the way something else.

ikedas:
Using [Unicode Collate algorthm](https://metacpan.org/pod/Unicode::Collate) may solve the problem better for European and other languages (Still only sorting by character code is possible for Kanji, though).

salaun-urennes1:
Thanks for suggesting this alternative.

It seems Unicode::Collate is installed on my server. Does it come with Perl?

Using this module for sorting would however require to write a custom sort subroutine to be used in TT2 templates...

ikedas:
Currently Unicode::Collate is not a core module of Perl, but we may be able to add it to our `cpanfile`, i.e. we may add it as one of dependency module of Sympa.  Of course we need to write any code to use this module in Sympa, but when at last we do, it may be used as easy as builtin sort().

It's midnight here.  See you in the next week!

ikedas:
The following opinion is separate from the pros and cons of this PR and may be continued later in a different location.

Currently, users need to distinguish slightly different titles in order to find the appropriate scenario. Sorting would make it easier to distinguish between similar titles by placing them closer to the differences. In other words, the fundamental problem may be that there are too many scenarios for each function.

For example, for the send scenario, it would be much easier to understand if several conditions were selected individually, such as:
- Who can post?
- Does the posting require the confirmation of the author?
- What to do if a post cannot be submitted: Moderation or rejection
- Authentication method: No authentication (smtp), confirmation (md5), S/MIME (smime)
- and so on.

All possible combinations of these conditions are made into separate scenarios, which means that we have to prepare many scenarios with similar titles.

salaun-urennes1:
Thank you for summarizing the problem.
I'm on off next week.
We can go on this discussion in May...

salaun-urennes1:
I agree with you comment; scenario titles are too heterogeneous. Given the number of scenarios, especially for `send`, it is hard for list owners to make a clever configuration choice for their list.

At our University we recently made a significant work to simplify the situation:
1. we analyzed all default and custom `send` scenarios to find out a) how many lists use each scenario and b) compare the behavior of each scenario. All these informations were summarized in a spreadsheet that looks like this 
image
2. we listed what `send` scenario is really worth maintaining (in our context). Others have been hidden using `:ignore` files;
3. we have renamed remaining `send` scenarios using a consistent schema. Here is an example: `send.ur1modereetrejetsinon:title.fr Owners/Editors: moderated ; staff: mederated ; other: reject`

ikedas:
@salaun-urennes1 , I was in the process of doing exactly the same thing (and had gavin up because I got fed up with the complexity)!
If you don't mind, please share the results of your work.

ikedas:
I convert this PR to Draft once.  Let's continue discuttion.

salaun-urennes1:
Here is the list of scenarios we ignore :
```
./default/scenari/send.owner:ignore
./default/scenari/send.private_smime:ignore
./default/scenari/send.newsletterkeyonly:ignore
./default/scenari/send.privateoreditorkey:ignore
./default/scenari/send.privateandnomultipartoreditorkey:ignore
./default/scenari/send.intranetorprivate:ignore
./default/scenari/send.privateandeditorkey:ignore
./default/scenari/send.private:ignore
./default/scenari/send.editorkey:ignore
./default/scenari/send.editorkeyonlyauth:ignore
./default/scenari/send.publicnomultipart:ignore
./default/scenari/send.default:ignore
./default/scenari/send.privateorpublickey:ignore
./default/scenari/send.public:ignore
./default/scenari/send.ownerauth:ignore
./default/scenari/send.editordkim:ignore
./default/scenari/send.publicnoattachment:ignore
./default/scenari/send.publickey:ignore
./default/scenari/send.newsletter:ignore
./default/scenari/send.public_nobcc:ignore
./default/scenari/send.privatekey:ignore
./default/scenari/send.privatekeyandeditorkeyonly:ignore
./default/scenari/send.closed:ignore
./default/scenari/send.intranet:ignore
./default/scenari/send.editorkeyonly:ignore
```

PR #1687, opened 2023-06-15
🐛 — Fix error when rejecting message from direct URL

ldidry:
When using the reject URL sent in the moderation notification mail, there is no message_template argument, creating an error. This commit adds a default value for the argument.

The reject URL is like `https://example.org/sympa/reject/the_list/token_created_by_sympa`.

The error was
```
DIED: Parameter $tpl is not defined at /var/sympa/share/sympa/lib/Sympa/Message/Template.pm line 64.
```

And a stacktrace, leading me to the FCGI script.


ikedas:
I agree to tag this pr as [ready].

racke:
I agree. Tagged it as ready and added it to the 6.2.74 milestone.

Issue #57, opened 2017-09-15
Support for LDAP paged queries


rpet:
I'm bringing here an enhancement request regarding LDAP paged queries already filed in this two tickets:

- https://sourcesup.renater.fr/tracker/?aid=3525 (opened 2007-11-18)
- https://sourcesup.renater.fr/tracker/?aid=7369  (opened 2010-12-17, duplicate of 3525, provides a patch made by leon.hedding@om.org)

(The bug tracker at **renater.fr** seems to be restricted to users from French institutions, so I'm not able to contribute)

The rationale behind this request was already documented in those tickets (I'm adding some opinions from my own)

 - As an LDAP administrator I'd prefer to use paged searches than raising the search result limit each time we think a query might be approaching that limit.
 - In large deployments it could not be feasible to raise the limit to any arbitrary value.
 - Paged searches allow for more escalabe deployments
 - As an LDAP user I might  not have full control over the settings of the LDAP server I have to use
 - It's perfectly fair to use LDAP paged seaches. By no means it's a way to "work around legitimate limitations" of any RFC, as it was estated by someone in https://sourcesup.renater.fr/tracker/?aid=3525

ikedas:
- For example on OpenLDAP, by default, page size limit seems equal to "hard size limit" (size limit of total count of results enforced by server), and not to be enforced by server (cf. [Admin manual](http://www.openldap.org/doc/admin24/limits.html#Size%20limits%20and%20Paged%20Results)).
- On Active Directory, page size limit is enforced by MaxPageSize LDAP policy of server, and its value is lower than MaxResultSetSize (size limit of total count of results enforced by server) (cf.
 [Doc on technet](https://technet.microsoft.com/en-us/library/dn743658.aspx)).

I guess MaxPageSize is the challenge on Active Directory, because its default is slightly low (1000) _and_ it is enforced by server.  If several other directory servers also have the same problem, I think introducing paged search seems useful.

bertaviader:
I was trying to find this link: https://sourcesup.renater.fr/tracker/?aid=7369  but it seems is down. Do you have the code? I need it to find a solution about pagination.
Thanks,
Berta

lchanouha:
Hello,
Instead of modifiying LDAP server soft limit, or di paged control (my hard limit is higher), I just added sizelimit arg on Sympa/List.pm
`
    $mesg = $db->do_operation(
        'search',
        base   => "$ldap_suffix",
        filter => "$ldap_filter",
[...]

PR #1733, opened 2023-11-01
Inclusion from LDAP data sources supports RFC 2696 Paged Results control (#57)

farialima:
Fix for #57

It includes a new configuration setting "page size" that's used both to enable paging, and to define the page size. I tried to integrate a description in the setting, as I think it's good to keep the documentation in/next to the functionality, but I'd suppose that the description/documentation can be improved, feedback really welcome on this.

Now, The code for the LDAP data sources is quite convoluted, difficult to read, etc... It took me a looong time to understand that "two-level" was practically working out of the box once I had implemented "one-level", LOL). I've the feeling that it could be simplified. I've entered a different ticket on this: #1738 , as this is clearly not part of #57 .

### DIFF
--- a/src/lib/Sympa/DataSource/LDAP.pm
+++ b/src/lib/Sympa/DataSource/LDAP.pm
@@ -68,20 +68,29 @@ sub _open_operation {
     my $ldap_attrs  = $options{attrs}  || $self->{attrs};
     my $ldap_scope  = $options{scope}  || $self->{scope};
 
-    my $mesg = $self->{_db}->do_operation(
-        'search',
+    my @args = (
         base   => $ldap_suffix,
         filter => $ldap_filter,
         attrs  => [split /\s*,\s*/, $ldap_attrs],
         scope  => $ldap_scope,
         control=> $self->{_page} ? [$self->{_page}] : []
     );
+
+    my $mesg = $self->{_db}->do_operation('search', @args);
+
     unless ($mesg) {
         $log->syslog(
             'err',
             'LDAP search (single level) failed: %s with data source %s',
             $self->{_db}->error, $self
         );
+
+        if ($self->{_page}) {
+            # We had an abnormal exit, so let the server know we do not want any more
+            $self->{_page}->size(0);
+            $self->do_operation('search', @args);
+        }
+
         return undef;
     }
 
@@ -166,9 +175,7 @@ sub _load_next {
     }
 
     if ($self->{_page}) {
-        my( $response ) = $mesg->control( LDAP_CONTROL_PAGED );
-        my $cookie = $response->cookie;
-
+        my $cookie = $mesg->control( LDAP_CONTROL_PAGED )->cookie;
         $self->{_page}->cookie( $cookie );
     }
 
--- a/src/lib/Sympa/DataSource/LDAP.pm
+++ b/src/lib/Sympa/DataSource/LDAP.pm
@@ -45,10 +45,11 @@ sub _open {
     return undef unless $db and $db->connect;
     $self->{_db} = $db;
 
-    if ($db->__dbh->root_dse->supported_control(
+    my $pagesize = $options{pagesize} || $self->{pagesize};
+    if ($pagesize and $db->__dbh->root_dse->supported_control(
             Net::LDAP::Constant::LDAP_CONTROL_PAGED()
         )) {
-        $self->{_page} = Net::LDAP::Control::Paged->new( size => 1000 );
+        $self->{_page} = Net::LDAP::Control::Paged->new( size => $pagesize );
     }
 
     my $mesg = $self->_open_operation(%options);
--- a/src/lib/Sympa/DataSource/LDAP.pm
+++ b/src/lib/Sympa/DataSource/LDAP.pm
@@ -78,6 +78,11 @@ sub _open_operation {
 
     my $mesg = $self->{_db}->do_operation('search', @args);
 
+    if ($self->{_page} and $mesg) {
+        my $cookie = $mesg->control( LDAP_CONTROL_PAGED )->cookie;
+        $self->{_page}->cookie( $cookie );
+    }
+
     unless ($mesg) {
         $log->syslog(
             'err',
@@ -172,11 +177,6 @@ sub _load_next {
 
             last if $ldap_select eq 'first';
         }
-    }
-
-    if ($self->{_page}) {
-        my $cookie = $mesg->control( LDAP_CONTROL_PAGED )->cookie;
-        $self->{_page}->cookie( $cookie );
     }
 
     return [@retrieved];
--- a/src/lib/Sympa/DataSource/LDAP.pm
+++ b/src/lib/Sympa/DataSource/LDAP.pm
@@ -25,6 +25,9 @@ package Sympa::DataSource::LDAP;
 use strict;
 use warnings;
 
+use Net::LDAP::Control::Paged;
+use Net::LDAP::Constant qw( LDAP_CONTROL_PAGED );
+
 use Sympa::Database;
 use Sympa::Log;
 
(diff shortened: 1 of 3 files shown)


src/lib/Sympa/DataSource/LDAP.pm
```diff
@@ -25,6 +25,9 @@ package Sympa::DataSource::LDAP;
 use strict;
 use warnings;
 
+use Net::LDAP::Control::Paged;
+use Net::LDAP::Constant qw( LDAP_CONTROL_PAGED );
```
ikedas, line 29: To avoid the namespace pollution, we'd be betterto refrain from using exported symbols. Use `Net::LDAP::Constant::LDAP_CONTROL_PAGED()`, although it is redundant.

src/lib/Sympa/DataSource/LDAP.pm
```diff
@@ -108,7 +125,13 @@ sub _load_next {
     }
 
     my @retrieved;
-    my $mesg = $self->__dsh;
+    my $mesg;
+    if ($self->{_page} and $self->{_page}->cookie) {
```
ikedas, line 129: The cookie can be a string `"0"` by which this condition fails.  See [an issue on 389DS](https://pagure.io/389-ds-base/issue/47610).

src/lib/Sympa/DataSource/LDAP.pm
```diff
@@ -108,7 +125,13 @@ sub _load_next {
...
+        $mesg = $self->_open_operation(%options);
```
ikedas, line 130: - Why do search operation again? (I don't know if this is a appropriate way to do it.)
- Maybe we could use callback.

src/lib/Sympa/DataSource/LDAP.pm
```diff
@@ -61,19 +68,29 @@ sub _open_operation {
...
+
     unless ($mesg) {
         $log->syslog(
             'err',
             'LDAP search (single level) failed: %s with data source %s',
             $self->{_db}->error, $self
         );
+
+        if ($self->{_page}) {
+            # We had an abnormal exit, so let the server know we do not want any more
```
ikedas, line 89: Good! :+1:

src/lib/Sympa/DataSource/LDAP.pm
farialima: > [kedas](https://github.com/ikedas) [4 days ago](https://github.com/sympa-community/sympa/pull/1733#discussion_r1381100811)
> Why do search operation again? (I don't know if this is a appropriate way to do it.)

Yes, calling "search" again is the correct way to do that -- see the example in https://metacpan.org/pod/Net::LDAP::Control::Paged 

> Maybe we could use callback.

Yes, we could use a callback, but it would be more changes, and it wouldn't be simpler, I think

ikedas: > > [kedas](https://github.com/ikedas) [4 days ago](https://github.com/sympa-community/sympa/pull/1733#discussion_r1381100811)
> > Why do search operation again? (I don't know if this is a appropriate way to do it.)
> 
> Yes, calling "search" again is the correct way to do that -- see the example in https://metacpan.org/pod/Net::LDAP::Control::Paged

For me, your code looks doing search operation twice for the first page. The example does not.

farialima: Good catch ! although this was not the case in the first version of the code you looked at.

Reverted in bb40df3436be28e22c580cdde5496c0c3063c7d9.

ikedas: I understand. Thank you.

ikedas:
> There are a few things I'm not 100% happy in my implementation:
> 
> * it is not implemented for the two-level LDAP data source, only for one-level. It would be more complex for the two-level (the current code is a little complex right now, it could/should be refactored first to get paging in place), and I don't have an environment where to test it.
>   I think it's OK to merge it only for LDAP one-level, as the main use case is Active Directory which has a "SizeLimit" that cannot be configured, AFAIK, or at least not easily, and thus is easier to make work on the client. But maybe the maintainer would disagree...

What is the cause of the inability to apply pagination to LDAP two-level data source?

I think it would be better to be able to do it because of less astonishments for users.

> * Ideally, the new code should be in DatabaseDriver::LDAP instead, but it would be more complex - database drivers don't control the returned dataset so a whole class would need to be implemented, I think. It would be much more code. So I think it's OK, even better, to have it in the DataSource::LDAP class

I agree.  In fact, it does not seem that pagination would be necessary for LDAP use by Sympa other than as a data source.

ikedas:
I got it.

`_load_next()` of `Sympa::DataSource::LDAP` may be called multiple times.  This behavior is useful for implementing two-level data source:  Multiple search operation may be executed repeatedly in single connection.

Your change for pagination uses this mechanism to repeatedly read multiple pages. As a result, it breaks the two-level data source.

I would like to make one suggestion. How about once I refactor the current code and change it to use callbacks, you can submit your changes to it again?

farialima:
> I got it.
> 
> `_load_next()` of `Sympa::DataSource::LDAP` may be called multiple times. This behavior is useful for implementing two-level data source: Multiple search operation may be executed repeatedly in single connection.
> 
> Your change for pagination uses this mechanism to repeatedly read multiple pages. As a result, it breaks the two-level data source.

Yes, exactly. 

I think I have it working, but I don't have a test setup yet for testing it... will take a few days.

> 
> @farialima , I would like to make one suggestion. How about once I refactor the current code and change it to use callbacks, you can submit your changes to it again?

I'm really not sure that using a callback improves things, but I will look at it.

farialima:
> > I got it.
> > `_load_next()` of `Sympa::DataSource::LDAP` may be called multiple times. This behavior is useful for implementing two-level data source: Multiple search operation may be executed repeatedly in single connection.
> > Your change for pagination uses this mechanism to repeatedly read multiple pages. As a result, it breaks the two-level data source.
> 
>> I think it [implementing the two-level LDAP data source] would be better to be able to do it because of less astonishments for users.
> 
> I think I have it working, but I don't have a test setup yet for testing it... will take a few days.
> 

now done - ready for review

> > @farialima , I would like to make one suggestion. How about once I refactor the current code and change it to use callbacks, you can submit your changes to it again?
> 
> I'm really not sure that using a callback improves things, but I will look at it.

I was indeed able to implemented two-level LDAP with very few changes, so I'd rather not refactor thing, because really the changes are minimal. But I've entered #1738 as a follow-up, I'll look at it

ikedas:
OK.  If we confirm that existing unit test for LDAP2 without pagination passes, i.e. at least the changes won't break existing feature, this PR may be merged.

ikedas:
@farialima , I have one more question.

Currently, unless the administrator sets `pagesize` explicitly, Sympa might not retrieve entire results from the server that enforces pagination.  How about the `pagesize` parameter has some positive default value? (I think this value would be better to be smaller than `MaxPageSize` setting of AD.)

farialima:
> @farialima , I have one more question.
> 
> Currently, unless the administrator sets `pagesize` explicitly, Sympa might not retrieve entire results from the server that enforces pagination. How about the `pagesize` parameter has some positive default value? (I think this value would be better to be smaller than `MaxPageSize` setting of AD.)

I've been hesitating indeed to automatically use paging if it's available. But I don't find a generic way to query the MaxPageSize of AD, so I couldn't get to decide what default value to use... I agree it's one more small complexity, but at least at first, I'd rather make it manual.. If you have any idea of how it could work, I'd be a taker, but I haven't found a way to do it, in a way I would find reliable enough...

ikedas:
I don't have a good idea either... All right, let's start with manual onfiguration first.

PR #1741, opened 2023-11-23
Upgrade command should fail if no previuos version number can be found.

woodfighter:
Currently the `sympa upgrade` cli command when not given an explicit previous version via the from parameter checks the file data_structure.version in the config directory. But instead of stopping if this also doesn't exist it assumes the previous version to be zero an consequently tries to apply all data upgrades. This patch just lets upgrade err on the safe side. If the previous behavior is needed passing along a `--from 0` should suffice.


ikedas:
Confirmed.  @woodfighter , thank you for fixing bug!  This PR will be merged in some days.

Issue #1752, opened 2023-11-29
Sympa Accessibility Issue - 22 - The items in the multiselect box are not readable fully.


rm1-uiuc:
The University of Illinois at Urbana-Champaign, Technology Services Accessibility Testing team conducted an Accessibility Review of Sympa on September 25, 2023 and found 26 issues that need to be fixed to make the interface accessible for visual impaired users.

Version
-------
6.2.72

Installation method
-------------------
Source Package

Expected behavior
-----------------
When a user increases the browser resolution to 200% items in the multiselect combo boxes should still be able to be read the content and it should not be caught off to be read by a screen reader.

Actual behavior
---------------
The items in the multiselect box are not readable fully. Items are cut off. When this is zoomed to 200%, the readable content gets lesser and not be read by a screen reader.

Steps to reproduce
------------------
Sympa Accessibility Issue - 22 was determined by manual functional testing performed by the user navigating with keyboard only or using assistive technology such as screen reader. See attached review for more info on specific location in interface.

The items in the multiselect box are not readable fully. Items are cut off. When this is zoomed to 200%, the readable content gets lesser.

[WCAG Level AA 1.4.4. Resize Text](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html)

![image](https://github.com/sympa-community/sympa/assets/144282683/cee9a449-ce35-4a78-ac87-3ada18ad6bec)

Fig 23 Text being cut off issue.

Additional information
----------------------
Attached is the Accessibility Review conducted on Sept 25,2023
[Tech Services Review - Accessibility Evaluation Report for Sympa.pdf](https://github.com/sympa-community/sympa/files/13505949/Tech.Services.Review.-.Accessibility.Evaluation.Report.for.Sympa.pdf)

ldidry:
I don’t know how it’s supposed to be handled but we can use this piece of css to make the option horizontally scrollable:
```css
select[multiple] option {
    overflow-x: scroll;
}
```

I did not found anything on the web allowing to break text on multiple lines in a `select[multiple]`  tag.

PR #1757, opened 2023-11-30
♿ — WWSympa: The items in the multiselect box are not readable fully (#1752)

ldidry:
I asked a friend who worked in web accessibility about [my solution](https://github.com/sympa-community/sympa/issues/1752#issuecomment-1833384255) and he said it was quite the best you can do about this specific problem.


default/web_tt2/css.tt2
ikedas, line 1: This looks working as expected.  Thanks @ldidry !

PR #1782, opened 2023-12-16
Use browser's built-in "date" widget instead of jQuery UI Datepicker that is inaccessible using keyboard (#1751)

### LINKED ISSUE #1751: Sympa Accessibility Issue - 21 - Unable to select date from the calendar picker using keyboard
The University of Illinois at Urbana-Champaign, Technology Services Accessibility Testing team conducted an Accessibility Review of Sympa on September 25, 2023 and found 26 issues that need to be fixed to make the interface accessible for visual impaired users.

Version
-------
6.2.72

Installation method
-------------------
Source Package

Expected behavior
-----------------
User should be able to select a date in the calendar picker using a keyboard

Actual behavior
---------------
Unable to select date from the calendar picker using keyboard. When navigating through keyboard, the calendar picker popups when hitting enter but tabbing into the picker makes the picker disappear.

Steps to reproduce
------------------
Sympa Accessibility Issue - 21 was determined by manual functional testing performed by the user navigating with keyboard only or using assistive technology such as screen reader. See attached review for more info on specific location in interface.

Unable to select date from the calendar picker using keyboard. When navigating through keyboard, the calendar picker popups when hitting enter but tabbing into the picker makes the picker disappear.

[2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)

image
Fig 22 Keyboard accessibility issue on date picker.

Additional information
----------------------
Attached is the Accessibility Review conducted on Sept 25,2023
[Tech Services Review - Accessibility Evaluation Report for Sympa.pdf](https://github.com/sympa-community/sympa/files/13505476/Tech.Services.Review.-.Accessibility.Evaluation.Report.for.Sympa.pdf)

ikedas:
Instead, use browser's built-in widget with <input type="date"> element.

This may fix #1751.

### DIFF
--- a/default/web_tt2/head_javascript.tt2
+++ b/default/web_tt2/head_javascript.tt2
@@ -11,18 +11,13 @@
 [%# A few configuration settings and miscellaneous vars. ~%]
 var sympa = {
     backText:           '[%"Back"|loc|escape_cstr%]',
-    calendarButtonText: '[%"Calendar"|loc|escape_cstr%]',
-    calendarFirstDay:   0,
     closeText:          '[%"Close"|loc|escape_cstr%]',
     copiedText:         '[%"Copied!"|loc|escape_cstr%]',
     copyPermalinkText:  '[%"Copy permalink"|loc|escape_cstr%]',
-    dayNames:           '[%"Sunday:Monday:Tuesday:Wednesday:Thursday:Friday:Saturday"|loc|escape_cstr%]'.split(":"),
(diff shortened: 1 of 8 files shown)


ldidry:
I don’t know if it would be useful for the few browsers which doesn’t support date input type, but maybe we could keep a `placeholder="yyyy-mm-dd"` on the input?

ikedas:
Done!

ldidry (approved):
LGTM

Issue #1797, opened 2024-02-15
Cross-robot include_sympa_list tries to include the wrong list, possibly failing with an imagined "inclusion loop"


dpoon:
If a mailing list <code><i>target-list</i>@<i>target-robot.example.edu</i></code> is configured with

<pre><code>include_sympa_list
name <i>blah</i>
listname <i>source-list</i>@<i>source-robot.example.edu</i></code></pre>

… then when Sympa performs the sync, it tries to include <code><i>source-list</i>@<i>target-robot.example.edu</i></code> instead.

Version
-------
6.2.66, but I believe the problematic code is the same in 6.2.72 and main.

Installation method
-------------------
Universe repo for Ubuntu 22.04: https://mirror.it.ubc.ca/ubuntu/pool/universe/s/sympa/sympa_6.2.66~dfsg-2_amd64.deb

Steps to reproduce
------------------
Configure mailing list data sources as described in the summary above (Admin → Edit List Config → Data Sources Setup → List inclusion).  Trigger a sync (<code>sympa.pl --sync_include=<i>target-list</i>@<i>target-robot.example.edu</i></code>).

Actual behavior
---------------
Sympa tries to load <i>source-list</i>@<i>target-robot.example.edu</i> instead of <i>source-list</i>@<i>source-robot.example.edu</i>.

Furthermore, if <i>source-list</i> and <i>target-list</i> share the same `listname` in different robots, then Sympa ends up trying to include <i>target-list</i>@<i>target-robot.example.edu</i>, causing the inclusion request to fail with

```
err main::#851 > Sympa::Spindle::spin#95 > Sympa::Request::Handler::include::_twist#202 > Sympa::Request::Handler::include::_update_users#370 > Sympa::DataSource::open#139 > Sympa::DataSource::List::_open#71 Loop detection in list inclusion: could not include again Sympa::DataSource::List <context=listname@target-robot.example.edu;id=4e750bd4;role=member;name=include_list listname@source-robot.example.edu> in list Sympa::List <listname@target-robot.example.edu>
```

… even though no such loop exists.

Expected behavior
-----------------
This used to work: Sympa would load <i>source-list</i>@<i>source-robot.example.edu</i>.

Additional information
----------------------
[...]

PR #1798, opened 2024-02-15
Fix cross-robot list inclusion (#1797)

dpoon:
If the Sympa::List->new($name, $robot, $options) constructor is called with an optional $robot parameter specified, then that $robot overrides any domain that may be present in $name.  When constructing a DataSource for inclusion, the robot of the DataSource should be inferred only if no domain is present in the list name.


benjyamon:
We ran into this bug and applied this fix to our production version 6.2.72 and it's fixed! So thanks and +1 to this pull request making it's way into a future release ;)

PR #1824, opened 2024-03-31
If lock fails, details should be included in the error message

ikedas:


racke:
That is really a good idea.

PR #1893, opened 2024-09-15
Update de.po

ilf:
"remove" was wrongly translated as "löschen", but that's "delete" - use correct translation "entfernen" instead


ikedas:
Hi @ilf ,
Please use the translation site https://translate.sympa.community/ to correct and/or improve the translation.  See also the [help](https://translate.sympa.community/pages/help/).

Thank you!

Issue #1235, opened 2021-09-09
Memory leak when Archive Download fails


dpc22:
Version
-------

6.2.64

Installation method
-------------------

My own rpm, derived from official source RPM. Redhat Enterprise Server 7.

Expected behavior
-----------------

wwsympa.pl should not leak 4 GBytes of memory when an archive download fails.

Actual behavior
---------------

(Output from Linux top command):
```
16817 sympa     20   0 4916564   4.2g   6612 S   0.0 26.8   4:30.05 wwsympa.fcgi                                                      
``` 
4.2g is 4.2 Gytes of Residient Set Size. That starts as about 200 MBytes.

Additional Information
---------------

One of my list owners attempted to download a 1.3 GBytes ZIP archive of their mail archives (about 3 to 4 GBytes of raw data). 

Apache timed out after (I think) 60 seconds: it is possible I need to increase the Apache default timeout. That left the wwsympa.fcgi process in question using 4.2 GBytes RSS, which does not appear to be returned to the system.

The list owner then repeated the process multiple times until *all* of the wwsympa processes were 4.2 GBytes in size.

I think that there might also be slow memory leaks in other Sympa daemons, but this largely explains why a virtual machine with 16 GBytes RAM was swapping like mad this morning until I restarted wwsympa. It normally only uses about 6 GBytes, with the rest spare/as buffer cache. OOM killer had stepped in, which is never a good sign on Linux.

I don't think that this memory leak happens if the download is successful. However I think  that I will disable list archive downloads for the time being...

dpc22:
Possibly worth adding: the Linux "zip" command can build these archives _much_ faster  than the Perl modules Sympa is using.

ikedas:
@dpc22 , can you check [this patch](https://github.com/sympa-community/sympa/compare/6.2.64...20f2558.patch)?
This fix requires a new module [Archive-Zip-SimpleZip](https://metacpan.org/pod/Archive::Zip::SimpleZip).

dpc22:
The patch applies, but  Archive-Zip-SimpleZip isn't available from the RHEL base or EPEL repositories, so I will need to install it using CPAN or such like. All of the other Perl dependencies for Sympa are available from EPEL. I'm also a bit dubious about : "Module version: 0.040", although I guess version numbers are fairly arbitrary.

[...]

PR #1902, opened 2024-10-18
Giving up Archive-Zip

ikedas:
Instead, make Archive-Zip-SimpleZip mandatory.

See also #1235.


xavierba:
This was needed for EL7 only. Now that EL7 is gone, this is fine from Fedora/EPEL side.

PR #1916, opened 2024-12-13
For HTML messages, add HTML tags to all URIs in message footer/header text

adam12b1:
Use URI::Find::Schemeless to add HTML tags around all URIs in message footer/header text, so links are clickable when text is appended to HTML messages. No effect on plaintext messages.

### DIFF
--- a/src/lib/Sympa/Message.pm
+++ b/src/lib/Sympa/Message.pm
@@ -63,6 +63,8 @@ use Sympa::Tools::Password;
 use Sympa::Tools::SMIME;
 use Sympa::Tools::Text;
 use Sympa::User;
+use CGI;
+use URI::Find::Schemeless;
 
 my $language = Sympa::Language->instance;
 my $log      = Sympa::Log->instance;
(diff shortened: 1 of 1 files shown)


src/lib/Sympa/Message.pm
```diff
@@ -63,6 +63,8 @@ use Sympa::Tools::Password;
 use Sympa::Tools::SMIME;
 use Sympa::Tools::Text;
 use Sympa::User;
+use CGI;
+use URI::Find::Schemeless;
```
ikedas, line 67: URI::Find::Schemeless should also be added to the `cpanfile` file on the top of source tree as a `requires` section.

src/lib/Sympa/Message.pm
```diff
@@ -2167,16 +2169,22 @@ sub _append_footer_header_to_part {
...
+        #---acb: and for all three header/footer parts, wrap all URIs
+        # in HTML tags, and keep newlines for readability and for
+        # outlook.com parsing:
         $header_msg = Sympa::Tools::Text::encode_html($header_msg);
+        $header_msg = uri_finder($header_msg);
         $header_msg =~ s/(\r\n|\r|\n)$//;       # strip the last newline.
-        $header_msg =~ s,(\r\n|\r|\n),<br/>,g;
+        $header_msg =~ s,(\r\n|\r|\n),<br/>$1,g; # keep newlines
         $footer_msg = Sympa::Tools::Text::encode_html($footer_msg);
+        $footer_msg = uri_finder($footer_msg);
```
ikedas, line 2180: HTML encoding (escaping) should be done only in `uri_finder()` below, i.e. `Sympa::Tools::Text::encode_html()` in above line should be removed. Or escaping will duplicate (and URIs including `&` will not be detected properly).

src/lib/Sympa/Message.pm
```diff
@@ -3700,6 +3708,26 @@ sub get_id {
...
+
+    # insert "mailto:" on all email addresses, so URI::Find will find them:
+    $text =~ s,(^|\s)(\S+)@(\S+)($|\s),$1mailto:$2\@$3$4,g;
+
+    # wrap all URIs in HTML tags
+    my $finder = URI::Find::Schemeless->new(sub {
+        my($uri, $orig_uri) = @_;
+        return qq|<a href="$uri">$orig_uri</a>|;
+                                            });
+    $finder->find(\$text, \&CGI::escapeHTML);
```
ikedas, line 3723: Use `Sympa::Tools::Text::escape_html()` instead of `CGI::encodeHTML()`. The reason is simply that Sympa consistently uses the former.

src/lib/Sympa/Message.pm
```diff
@@ -3700,6 +3708,26 @@ sub get_id {
...
+
+    # and finally clean up the displayed link text on mailto links:
+    $text =~ s/>mailto:/>/g;
```
ikedas, line 3726: I personally think that this “`mailto:`” is included in the original text and may not be removed.

ikedas:
@adam12b1 , are you willing to update your PR?

cpanfile
```diff
@@ -166,6 +166,9 @@ requires 'URI::Escape', '>= 3.28';
 # libxml2-devel is needed to build the Perl code
 requires 'XML::LibXML', '>= 1.70';
 
+# Used for adding HTML tags to URI in message footers/headers:
+requires 'URI::Find::Schemeless, '20160806';
```
ikedas: A "'" is missing.

adam12b1: That was stupid of me, thank you.

src/lib/Sympa/Message.pm
ikedas: Please feel free to submit a new PR until you're satisfied!

I’ll play with this PR this weekend.

adam12b1:
@ikedas I've finally tested and I think your improvement looks good. It misses some cases (eg. an email address in quotes like "adam@electricembers.coop" will not get a link now) but that is probably fine - we can let that go in exchange for simpler code. So, let me know if there's anything else I should do to help get this PR resolved!

Issue #1918, opened 2024-12-24
automatic process does not sync owner and editor


simonelbaz:
Hi,

Version
-------
Reproduced in 6.2.70
The code is still the same in main branch

Installation method
-------------------
deb, bullseye distribution

Expected behavior
-----------------
owner and editor should be synced if defined in the config.tt2

Actual behavior
---------------
only members are synced by automatic process

Steps to reproduce
------------------
TT2 attached

[sympa-families-groupes-config.tt2.j2.txt](https://github.com/user-attachments/files/18240484/sympa-families-groupes-config.tt2.j2.txt)

Additional information
----------------------
Logs:

```
root@ppr1-sympa02-r1:/opt/docker/compose/sympa-docker# docker compose logs sympa-web-11-automatic | grep role
sympa-web-11-automatic-1  | notice Sympa::Spindle::ProcessRequest::_twist() Processing Sympa::Request [<action=include;context=xxxxxx;role=member>](mailto:action=include;context=xxxxxxxx;role=member)
sympa-web-11-automatic-1  | info Sympa::Request::Handler::include::_twist() Sympa::DataSource::LDAP [<context=xxxxxxx;id=0a62f42a;role=member;name=ldap>](mailto:context=xxxxxx;id=0a62f42a;role=member;name=ldap): 4 included, 0 deleted, 0 updated, 0 kept
sympa-web-11-automatic-1  | info Sympa::Request::Handler::include::_twist() Sympa::Request [<action=include;context=xxxxxxx;role=member>](mailto:action=include;context=xxxxxxxx;role=member): Success, 4 added, 0 deleted, 0 updated
sympa-web-11-automatic-1  | notice Sympa::Spindle::ProcessRequest::_twist() Processing Sympa::Request [<action=include;context=xxxxxx;role=member>](mailto:action=include;context=xxxxxxx;role=member)
sympa-web-11-automatic-1  | info Sympa::Request::Handler::include::_twist() Sympa::DataSource::LDAP [<context=xxxxxx;id=0a62f42a;role=member;name=ldap>](mailto:context=xxxxx;id=0a62f42a;role=member;name=ldap): 0 included, 0 deleted, 0 updated, 4 kept
sympa-web-11-automatic-1  | info Sympa::Request::Handler::include::_twist() Sympa::Request [<action=include;context=xxxxxxxx;role=member>](mailto:action=include;context=xxxxxx;role=member): Success, 0 added, 0 deleted, 0 updated
[...]

PR #1919, opened 2024-12-30
automatic syncs owner and editor at creation

simonelbaz:


simonelbaz:
Related https://github.com/sympa-community/sympa/issues/1918

PR #1995, opened 2025-09-15
[Web interface] add workaround for the popup ephemeralMsg blocking the view when not having JS

user3287:
**Not tested**, but idea should work

Possible good solutions:
- A closing button (❌) on the popup that also works without JS, which is AFAIK possible
- Make the popup a static part of the page
- Put the popup somewhere where it is not in the way of things

### DIFF
--- a/default/web_tt2/css.tt2
+++ b/default/web_tt2/css.tt2
@@ -1341,6 +1341,10 @@ h7 strong {
     margin-bottom: 1.5em;
     background-color: #fff;
     z-index:599;
+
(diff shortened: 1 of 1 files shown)


ikedas:
Hi @user3287 ,

> A closing button (❌) on the popup that also works without JS, which is AFAIK possible

See also:

- [Toggling elements with a hidden checkbox](https://developer.mozilla.org/en-US/docs/Web/CSS/:checked#toggling_elements_with_a_hidden_checkbox), MDN.

PR #2053, opened 2026-05-27
add reCAPTCHA v2 to subscribe form

RolinaS:
Protection anti spam
image


ikedas:
Hi @RolinaS ,

This looks a useful effort, but it is not available in all environments.  So it does not seem suitable for inclusion in the Sympa distribution.

Rather than providing a patch, I think it would be better to publish documentation explaining how to customize Sympa to use reCAPTCHA on each installation.

RolinaS:
Hi @ILESANMI-Olatunde,

Thank you for your feedback! I completely understand your concern about environments where Google reCAPTCHA may not be available.

I will close this pull request and have created a new contribution in the [sympa-contribs](https://github.com/sympa-community/sympa-contribs) repository instead, which seems more appropriate for this kind of customization: https://github.com/RolinaS/sympa-contribs/tree/feature/recaptcha-subscribe

It includes a documentation guide explaining how administrators can add a reCAPTCHA widget to the subscription form on their own installation, without modifying the Sympa source code.

Thanks again for pointing me in the right direction!

Best regards

PR #2071, opened 2026-08-18
Allow listmaster to modify recipient on resend

immae:
This commit is a change in the archive management, allowing the listmaster to chose a different e-mail as recipient when resending an e-mail.
The recipient is flexible (it doesn’t restrain to existing subscribers) on purpose: the use case that provoked this change was to submit a spam false positive to a remote mailer, which required sending the e-mail "exactly as is was sent initially"

### DIFF
--- a/default/mhonarc_rc.tt2
+++ b/default/mhonarc_rc.tt2
@@ -610,7 +610,12 @@ $POWERED_BY$
                         </label>
                 <% END %>
                 <input class="MainMenuLinks" type="submit" name="action_compose_mail" value="<%|loc%>Reply<%END%>" /><br />
+                <% IF is_listmaster %>
+                <input class="MainMenuLinks" type="submit" name="action_send_me" value="Re-deliver to:" />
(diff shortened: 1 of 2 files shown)


ikedas:
Hi @immae ,
Listmasters can switch their identity to that of any user. Then they may resend the message to that recipient.

immae:
> Hi @immae , Listmasters can switch their identity to that of any user. Then they may resend the message to that recipient.

Noted, but it doesn’t solve my usecase: I wanted to send the e-mail to an adress that didn’t subscribe to the mailing-list (in this case, some e-mail provider who wanted to analyse why the e-mail was considered as a spam by their system, they wanted the exact e-mail re-sent by the same system)

ikedas:
In your use case, you'd be better to retrieve (copy) the message in question from the server and to send it as data (an attachment) to the intended recipient. If you use a mechanism such as message forwarding, some parts of the message including header fields may be rewritten or deleted, thus accuracy would be compromised.

Please see also the description on the [archive structure](https://www.sympa.community/manual/customize/archives.html#archives-structure) in the documentation.

PR #2074, opened 2026-08-26
Log DMARC DNS lookup failures instead of silently skipping protection

paulmenzel:


paulmenzel:
Description is in the commit message.

